PR #1, opened 2023-06-01
SMP cleanup

krystian-hebel:


xen/arch/x86/smpboot.c
```diff
@@ -77,15 +77,18 @@ struct cpuinfo_x86 cpu_data[NR_CPUS] =
...
+    CPU_STATE_INIT,     /* slave -> master: Early bringup phase 1 completed */
+    CPU_STATE_CALLOUT,  /* master -> slave: Start early bringup phase 2 */
     CPU_STATE_CALLIN,   /* slave -> master: Completed phase 2 */
     CPU_STATE_ONLINE    /* master -> slave: Go fully online now. */
-} cpu_state;
-#define set_cpu_state(state) do { smp_mb(); cpu_state = (state); } while (0)
+};
+#define set_cpu_state(cpu, state) do { \
+    smp_mb(); \
+    cpu_data[cpu].cpu_state = (state); \
```
SergiiDmytruk, line 90: Would adding parenthesis around `cpu` make any difference in this case?

xen/arch/x86/smpboot.c
```diff
@@ -641,34 +642,40 @@ static int do_boot_cpu(int apicid, int cpu)
...
+        for ( timeout = 0; timeout < 20000; timeout++ )
+        {
+            if ( cpu_data[cpu].cpu_state == CPU_STATE_INIT )
+                break;
+            udelay(100);
+        }
+
         /* Allow AP to start initializing. */
-        set_cpu_state(CPU_STATE_CALLOUT);
+        set_cpu_state(cpu, CPU_STATE_CALLOUT);
```
SergiiDmytruk: Shouldn't we check for `cpu_data[cpu].cpu_state == CPU_STATE_INIT` before changing the state to catch timeout of the loop above?

SergiiDmytruk (changes requested):
I think a timeout check is missing, otherwise looks good.

However, mind that I'm seeing this code for the first time and don't know how the changes fit into the larger code base.

xen/arch/x86/smpboot.c
krystian-hebel: It doesn't, index is always treated as integer expression (or pointer expression, index and array name can be swapped because coding in C is fun). I've left it without parenthesis.

xen/arch/x86/smpboot.c
krystian-hebel: Right, I forgot to add it back after rebasing. Originally I tested for timeout value, but by checking state the timeout value doesn't have to be repeated in the code. This makes it easier to adjust the value, if needed, at the cost of one additional memory access.

SergiiDmytruk (approved):

xen/arch/x86/setup.c
```diff
@@ -380,9 +380,9 @@ static void __init normalise_cpu_order(void)
 
         /* Switch the best-matching CPU with the next CPU in logical order. */
         j = cpumask_next(i, &cpu_present_map);
-        apicid = x86_cpu_to_apicid[min_cpu];
-        x86_cpu_to_apicid[min_cpu] = x86_cpu_to_apicid[j];
-        x86_cpu_to_apicid[j] = apicid;
+        apicid = cpu_physical_id(min_cpu);
+        cpu_physical_id(min_cpu) = cpu_physical_id(j);
```
miczyg1, line 384: This notation looks quite weird, i.e. it looks like a function call on both sides of the equation, which may be confusing.

If it was CPU_PHYSICAL_ID(x) then it would probably look more intuitive.

xen/arch/x86/smpboot.c
```diff
@@ -322,41 +315,24 @@ static void set_cpu_sibling_map(unsigned int cpu)
...
-    if ( sl_status ) {
-        uint64_t misc_enable;
-        uint32_t my_apicid;
-        struct txt_sinit_mle_data *sinit_mle =
-              txt_sinit_mle_data_start(__va(read_txt_reg(TXTCR_HEAP_BASE)));
-
-        /* TXT released us with MONITOR disabled in IA32_MISC_ENABLE. */
-        rdmsrl(MSR_IA32_MISC_ENABLE, misc_enable);
-        wrmsrl(MSR_IA32_MISC_ENABLE,
-               misc_enable | MSR_IA32_MISC_ENABLE_MONITOR_ENABLE);
```
miczyg1, line 344: No longer needed? Why?

xen/arch/x86/smpboot.c
```diff
@@ -322,41 +315,24 @@ static void set_cpu_sibling_map(unsigned int cpu)
...
-
-        /* get_apic_id() reads from x2APIC if it thinks it is enabled. */
-        x2apic_ap_setup();
-        my_apicid = get_apic_id();
-
-        while ( my_apicid != x86_cpu_to_apicid[cpu] ) {
-            asm volatile ("monitor; xor %0,%0; mwait"
-                          :: "a"(__va(sinit_mle->rlp_wakeup_addr)), "c"(0),
-                          "d"(0) : "memory");
```
miczyg1, line 353: No longer needed? Why?

miczyg1 (changes requested):

xen/arch/x86/setup.c
krystian-hebel: I followed style used in Xen, e.g. https://github.com/TrenchBoot/xen/blob/master/xen/arch/x86/include/asm/percpu.h or https://github.com/TrenchBoot/xen/blob/master/xen/arch/x86/irq.c#L1083.

xen/arch/x86/smpboot.c
krystian-hebel: `MSR_IA32_MISC_ENABLE` for APs is written with value read from BSP in one of further initialization steps. This had to be done earlier to stop APs from proceeding, but it's no longer the case, as described [in commit message](https://github.com/TrenchBoot/xen/pull/1/commits/675f98090bc9ca835cf338999edef5258adcc251).

xen/arch/x86/smpboot.c
krystian-hebel: This was done to hold APs here until all notifiers are called. Now they just loop on `while ( cpu_data[cpu].cpu_state != CPU_STATE_CALLOUT )`, both for TXT and non-TXT path.

xen/arch/x86/setup.c
miczyg1: Oh well... Then nothing we can do about it.

miczyg1 (approved):

PR #3, opened 2023-06-28
Early-boot-time TPM2 support

SergiiDmytruk:


xen/arch/x86/sha256.c
```diff
@@ -0,0 +1,248 @@
...
+#define STATE_LEN 8  // In words
+
+static void sha256_compress(const uint8_t block[static 64], uint32_t state[static 8]) {
+	#define ROTR32(x, n)  (((0U + (x)) << (32 - (n))) | ((x) >> (n)))  // Assumes that x is uint32_t and 0 < n < 32
+
+	#define LOADSCHEDULE(i)  \
+		schedule[i] = (uint32_t)block[i * 4 + 0] << 24  \
+		            | (uint32_t)block[i * 4 + 1] << 16  \
+		            | (uint32_t)block[i * 4 + 2] <<  8  \
+		            | (uint32_t)block[i * 4 + 3] <<  0;
```
krystian-hebel: I was under the impression that Xen required function-like macros to skip the trailing semicolon, but I can't find anything in `CODING_STYLE`. Let's leave it as it was in the original file for now, but we may have to fix it later if upstream objects to this style. They may want us to add it to `docs/misra/exclude-list.json` as well.

xen/arch/x86/tpm.c
```diff
@@ -383,6 +397,338 @@ static void *create_log_event12(struct txt_ev_log_container_12 *evt_log,
...
+    uint8_t data[0];
+} __packed;
+
+struct tpm2_sequence_complete_rsp {
+    struct tpm_rsp_hdr h;
+    uint32_t paramSize;
+    uint32_t hashCount;
+    uint8_t hashes[0];
+    /*
+     * Each hashe is represented as:
```
krystian-hebel:
```suggestion
     * Each hash is represented as:
```

xen/arch/x86/tpm.c
```diff
@@ -383,6 +397,338 @@ static void *create_log_event12(struct txt_ev_log_container_12 *evt_log,
...
+#define TPM_ST_SESSIONS     0x8002
+
+#define TPM_ALG_SHA1        0x0004
+#define TPM_ALG_SHA256      0x000b
+#define TPM_ALG_NULL        0x0010
+
+#define TPM2_PCR_Extend                0x00000182
+#define TPM2_PCR_HashSequenceStart     0x00000186
+#define TPM2_PCR_SequenceUpdate        0x0000015C
+#define TPM2_PCR_EventSequenceCompete  0x00000185
```
krystian-hebel:
```suggestion
#define TPM2_PCR_EventSequenceComplete 0x00000185
```

xen/arch/x86/tpm.c
```diff
@@ -383,6 +397,338 @@ static void *create_log_event12(struct txt_ev_log_container_12 *evt_log,
...
+        }
+
+        size -= max_bytes;
+        buf += max_bytes;
+    }
+
+    cmd_rsp.finish_c = (struct tpm2_sequence_complete_cmd) {
+        .h.tag = swap16(TPM_ST_SESSIONS),
+        .h.paramSize = swap32(sizeof(cmd_rsp.finish_c) + size),
+        .h.ordinal = swap32(TPM2_PCR_EventSequenceCompete),
```
krystian-hebel:
```suggestion
        .h.ordinal = swap32(TPM2_PCR_EventSequenceComplete),
```

xen/arch/x86/tpm.c
```diff
@@ -383,6 +397,338 @@ static void *create_log_event12(struct txt_ev_log_container_12 *evt_log,
...
+    }
+
+    cmd_rsp.extend_c.hashCount = swap32(hash_count);
+
+    /* Fill in command size (size of the whole buffer). */
+    cmd_rsp.extend_c.h.paramSize = swap32(sizeof(cmd_rsp.extend_c) +
+                                          (p - cmd_rsp.extend_c.hashes)),
+
+    o_size = sizeof(cmd_rsp);
+    printk("o_size = %d\n", o_size);
```
krystian-hebel: These should be removed from the final version, although this is exactly what I would add for testing if it weren't here :wink:

xen/arch/x86/tpm.c
```diff
@@ -383,6 +397,338 @@ static void *create_log_event12(struct txt_ev_log_container_12 *evt_log,
...
+         * implement TPM2_GetCapability(). */
+        rc = tpm20_pcr_extend(loc, buf, /*size=*/0, /*pcr_handle=*/TPM_RH_NULL,
+                              test_sha1_digest, /*sha256_digest=*/NULL);
+        has_sha1 = (rc == 0);
+    }
+
+    if (has_sha1)
+        sha1_hash(buf, size, (uint32_t *)sha1_digest);
+    else
+        sha1_digest = NULL;
```
krystian-hebel: This will have to be a bit more complicated for event logging. I know this will be a separate PR, but I'll leave the comment here. You don't have to fix it here, but this is something to keep in mind.

Event log isn't initialized by Xen, it has to follow whatever format already exists. The header has information about algorithms contained inside, so information about SHA1 availability should come from outside of this function, assuming the header reflects what TPM implements, of course. I don't know how ACM processes it, in particular if it creates SHA1 entries in log if it isn't available on TPM (apparently there are already PTTs that allow to turn SHA1 off). So there are 4 possibilities:
- both TPM and event log support SHA1 -> no problem, hash it.
- both TPM and event log don't support SHA1 -> no problem, don't hash it.
- event log supports, TPM doesn't -> we probably need to write something to the log, either all 00s, FFs or some distinguishable magic value so anyone reading the log will notice that something is not right. We can't trust that event log is already zeroed, that would require trusting pre-DL event code.
- TPM supports, event log doesn't -> we can extend the PCR but not log it, or skip extending SHA1. In any case, PCR value can't be replayed, and we can't really check whether ACM extended this bank or not, since TPM will extend all banks for PCR-17 with DCE's hash anyway. I don't see any value with extending SHA1 in this case, but maybe @dpsmith can think of something.

xen/arch/x86/tpm.c
```diff
@@ -340,8 +355,7 @@ static void tpm_hash_extend12(unsigned loc, uint8_t *buf, unsigned size,
         .pcrNum = swap32(pcr),
     };
 
-    if ( out_digest != NULL )
```
krystian-hebel, line 343: Why is this removed? `out_digest` could be NULL if we ever run out of space in the event log.

krystian-hebel (commented):
I haven't tested it on the hardware yet, will do it later today, I hope.

xen/arch/x86/sha256.c
krystian-hebel: Feel free to mark this as resolved after reading, this is just informational.

xen/arch/x86/tpm.c
SergiiDmytruk: Because `sha1_hash(buf, size, (uint32_t *)out_digest);` above would cause a SEGFAULT before this check runs :)  So I concluded the check was left here accidentally.

xen/arch/x86/tpm.c
SergiiDmytruk: > event log supports, TPM doesn't -> we probably need to write something to the log, either all 00s, FFs or some distinguishable magic value so anyone reading the log will notice that something is not right. We can't trust that event log is already zeroed, that would require trusting pre-DL event code.

TXT spec has this:

```
Unsupported bank capping value

OneDigest is extended into PCR bank using an algorithm that is not supported by SINIT SW.
```

```
OneDigest is sequence of bytes: 01,00,00…00 with a total size equal to HashAlgId_DIGEST_SIZE
of capped TPM bank
```

krystian-hebel: We can use this value, although this is the opposite situation (TPM supports, SINIT doesn't).

krystian-hebel (approved):

PR #4, opened 2023-06-30
Event log for TPM2.0

SergiiDmytruk:
On top of PR #3.

Couldn't test this beyond building (QEMU/KVM refuses to passthrough host's SMX instructions and I don't have TPM to test it on hardware).

### DIFF
--- a/xen/arch/x86/tpm.c
+++ b/xen/arch/x86/tpm.c
@@ -287,7 +287,7 @@ union cmd_rsp {
     uint8_t buf[CMD_RSP_BUF_SIZE];
 };
 
-static void tpm_hash_extend12(unsigned loc, uint8_t *buf, unsigned size,
+static void tpm12_hash_extend(unsigned loc, uint8_t *buf, unsigned size,
                               unsigned pcr, uint8_t *out_digest)
 {
     union cmd_rsp cmd_rsp;
@@ -361,7 +361,7 @@ union cmd_rsp {
     struct extend_rsp extend_r;
 };
 
-static void tpm_hash_extend12(unsigned loc, uint8_t *buf, unsigned size,
+static void tpm12_hash_extend(unsigned loc, uint8_t *buf, unsigned size,
                               unsigned pcr, uint8_t *out_digest)
 {
     union cmd_rsp cmd_rsp;
@@ -598,9 +598,9 @@ union tpm2_cmd_rsp {
     struct tpm2_sequence_complete_rsp finish_r;
 };
 
-static uint32_t tpm_hash_extend20(unsigned loc, uint8_t *buf, unsigned size,
-                                  unsigned pcr,
-                                  struct tpm2_log_hashes *log_hashes)
+static uint32_t tpm2_hash_extend(unsigned loc, uint8_t *buf, unsigned size,
+                                 unsigned pcr,
+                                 struct tpm2_log_hashes *log_hashes)
 {
     uint32_t seq_handle;
     unsigned max_bytes = MAX_HASH_BLOCK;
@@ -786,9 +786,9 @@ static bool tpm_supports_hash(unsigned loc, const struct tpm2_log_hash *hash)
     return rc == 0;
 }
 
-static uint32_t tpm_hash_extend20(unsigned loc, uint8_t *buf, unsigned size,
-                                  unsigned pcr,
-                                  const struct tpm2_log_hashes *log_hashes)
+static uint32_t tpm2_hash_extend(unsigned loc, uint8_t *buf, unsigned size,
+                                 unsigned pcr,
+                                 const struct tpm2_log_hashes *log_hashes)
 {
     uint32_t rc;
     unsigned i;
@@ -954,7 +954,7 @@ void tpm_hash_extend(unsigned loc, unsigned pcr, uint8_t *buf, unsigned size,
         if ( entry_digest == NULL )
             entry_digest = sha1_digest;
 
-        tpm_hash_extend12(loc, buf, size, pcr, entry_digest);
+        tpm12_hash_extend(loc, buf, size, pcr, entry_digest);
     } else {
         uint32_t rc;
 
@@ -963,7 +963,7 @@ void tpm_hash_extend(unsigned loc, unsigned pcr, uint8_t *buf, unsigned size,
             create_log_event20(evt_log, evt_log_size, pcr, type, log_data,
                                log_data_size);
 
-        rc = tpm_hash_extend20(loc, buf, size, pcr, &log_hashes);
+        rc = tpm2_hash_extend(loc, buf, size, pcr, &log_hashes);
         if ( rc != 0 ) {
 #ifndef __EARLY_TPM__
             printk(XENLOG_ERR "Extending PCR%u failed with TPM error: 0x%08x\n",
--- a/xen/arch/x86/tpm.c
+++ b/xen/arch/x86/tpm.c
@@ -34,6 +34,16 @@ asm (
 #endif
 #define __va(x)     _p(x)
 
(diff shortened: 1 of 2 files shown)


xen/arch/x86/sha256.c
miczyg1, line 1: SHA256 is not x86 specific. It should probably be placed in more common place.

xen/arch/x86/tpm.c
```diff
@@ -34,12 +34,22 @@ asm (
 #endif
 #define __va(x)     _p(x)
 
+void *memset(void *dest, int c, size_t n)
```
miczyg1, line 37: Isn't any memset available in Xen?

xen/arch/x86/tpm.c
```diff
@@ -375,14 +399,520 @@ static void *create_log_event12(struct txt_ev_log_container_12 *evt_log,
...
+    struct tpm_rsp_hdr r;
+    struct tpm2_sequence_start_cmd start_c;
+    struct tpm2_sequence_start_rsp start_r;
+    struct tpm2_sequence_update_cmd update_c;
+    struct tpm2_sequence_update_rsp update_r;
+    struct tpm2_sequence_complete_cmd finish_c;
+    struct tpm2_sequence_complete_rsp finish_r;
+};
+
+static uint32_t tpm_hash_extend20(unsigned loc, uint8_t *buf, unsigned size,
```
miczyg1: Can we use `tpm2_` instead of `20` as a suffix? Most of the name start with tpm2

xen/arch/x86/sha256.c
SergiiDmytruk: Then SHA1 code which is already on the branch should also be moved, I added SHA256 in an analogous way.  (By the way, this is from #3.)

xen/arch/x86/tpm.c
SergiiDmytruk: It's not available when this file is compiled separately with `__EARLY_TPM__`.

xen/arch/x86/tpm.c
SergiiDmytruk: TPM1 version is called `tpm_hash_extend12`, so need to probably rename both at the same time or leave as is.

xen/arch/x86/sha256.c
miczyg1: Hmm then why it is not targeted at #3 ? How to not review changes two times here and in #3 ?

xen/arch/x86/tpm.c
miczyg1: I see.

xen/arch/x86/tpm.c
miczyg1: or have `tpm12_hash_extend` and `tpm2_hash_extend`. I assume we already use `tpm_hash_extend` as a version-agnostic higher level API?

xen/arch/x86/sha256.c
SergiiDmytruk: The expectation was that #3 will be merged by the time this is reviewed.  Changed target.

xen/arch/x86/tpm.c
SergiiDmytruk: > I assume we already use tpm_hash_extend as a version-agnostic higher level API?

Yes.  Renamed.

xen/arch/x86/sha256.c
SergiiDmytruk: Moved to `xen/lib` which already has xxhash.

xen/arch/x86/tpm.c
```diff
@@ -703,48 +771,159 @@ static uint32_t tpm20_pcr_extend(unsigned loc, uint32_t pcr_handle,
...
+                   hash->alg);
+            continue;
+        }
 
-    sha256_hash(buf, size, (uint32_t *)sha256_digest);
+        if ( hash->alg == TPM_ALG_SHA1 )
+            sha1_hash(buf, size, (uint32_t *)hash->data);
+        else if ( hash->alg == TPM_ALG_SHA256 )
+            sha256_hash(buf, size, (uint32_t *)hash->data);
+        else
```
krystian-hebel, line 813: This `else` doesn't have a statement so it is collapsed with following `if`, this probably isn't intended.

xen/arch/x86/tpm.c
```diff
@@ -703,48 +771,159 @@ static uint32_t tpm20_pcr_extend(unsigned loc, uint32_t pcr_handle,
...
     relinquish_locality(loc);
 
     return rc;
 }
 
 #endif /* __EARLY_TPM__ */
 
+static struct heap_event_log_pointer_element2_1 *find_evt_log_ext_data(void)
+{
+    struct txt_sinit_mle_data *sinit_mle;
```
krystian-hebel: `EVENT_LOG_POINTER2_1` lives in OS-SINIT data, it isn't moved to SINIT-MLE data.

I think I know where this error comes from - in TXT Software Development Guide all extended elements are listed in SINIT-MLE section, even though not all of the are used there.

xen/arch/x86/tpm.c
```diff
@@ -703,48 +771,159 @@ static uint32_t tpm20_pcr_extend(unsigned loc, uint32_t pcr_handle,
...
+
+    return (void *)&ext_data->data[0];
+}
+
+static struct tpm2_log_hashes
+create_log_event20(struct tpm2_spec_id_event *evt_log, uint32_t evt_log_size,
+                   uint32_t pcr, uint32_t type, uint8_t *data,
+                   unsigned data_size)
+{
+    struct tpm2_log_hashes log_hashes = {0};
```
krystian-hebel, line 863: I think it would be safer to allocate this struct on caller's side and pass pointer to it as an argument. It works for now, but if for any reason (e.g. different optimization settings) this function wouldn't be inlined, compiler may choose to return this as a pointer to structure on called function's stack, which may be overwritten.

krystian-hebel (changes requested):

xen/arch/x86/tpm.c
SergiiDmytruk: Compiler optimizations must not change behaviour of the code in any way, they definitely shouldn't completely break  correct code, otherwise it's a compiler bug.

Returning structs is always implemented by allocating object on the caller side and passing pointer to the callee.  When callee's body is known, copying from local variable to result can be elided by writing directly into the result, which is an optimization.  So I don't believe this requires any changes, returning structs by value works fine in C.

xen/arch/x86/tpm.c
SergiiDmytruk: It doesn't collapse with that `if`, there is a semicolon after comment  on the next line.

xen/arch/x86/tpm.c
SergiiDmytruk: > I think I know where this error comes from - in TXT Software Development Guide all extended elements are listed in SINIT-MLE section, even though not all of the are used there.

Yes, I might have not realized that extended data appears in two different structures with identically named fields in the spec.  Thanks.

xen/arch/x86/tpm.c
krystian-hebel: Right, missed that because of different color scheme than I'm used to.

xen/arch/x86/tpm.c
krystian-hebel: You are correct, for some reason I thought that a pointer was returned instead of whole struct.

Just for completeness of discussion, structs can also be returned in registers "if they fit", but what that exactly means depends on compiler, calling convention and additional switches used. Nevertheless, this won't be an issue here, both because the structure is relatively big and because both caller and called function are in the same compile unit, so they always use the same way of passing data between each other.

krystian-hebel (approved):

PR #7, opened 2023-10-28
Switch to using SLRT interface and process DRTM policy

SergiiDmytruk:


xen/arch/x86/tpm.c
```diff
@@ -978,11 +971,150 @@ void tpm_hash_extend(unsigned loc, unsigned pcr, uint8_t *buf, unsigned size,
...
+    policy_entry = (struct slr_policy_entry *)
+        ((uint8_t *)policy + sizeof(*policy));
+    tpm_hash_extend(DRTM_LOC, policy_entry->pcr, (uint8_t *)slrt,
+                    policy_entry->size, TXT_EVTYPE_SLRT, NULL, 0);
+}
+
+void __stdcall tpm_early_extend(uint32_t *mbi)
+{
+    tpm_extend_mbi(mbi);
+    tpm_extend_slrt();
```
krystian-hebel: Do we need to do it now? Xen doesn't use it until much later, when we have ways of communicating errors to users, as well as much faster hashing functions.

xen/arch/x86/tpm.c
```diff
@@ -978,11 +971,150 @@ void tpm_hash_extend(unsigned loc, unsigned pcr, uint8_t *buf, unsigned size,
...
+    if ( slr_next_entry_by_tag(slrt, NULL, SLR_ENTRY_INTEL_INFO) == NULL )
+        TPM_PANIC("SLRT is missing Intel-specific information!\n");
+
+    policy = (struct slr_entry_policy *)
+        slr_next_entry_by_tag(slrt, NULL, SLR_ENTRY_DRTM_POLICY);
+    if (policy == NULL)
+        TPM_PANIC("SLRT is missing DRTM policy!\n");
+
+    /* SLRT policy entry must be the first one (so measuring order matches
+     * policy order) and point to SLRT. */
```
krystian-hebel: In this case, first measurement (except for "internal" TXT stuff) is that of MBI, so it should be first to keep the relative order, even if this is done blindly before SLRT is parsed. This may require additional changes on GRUB side in https://github.com/TrenchBoot/grub/pull/14

xen/arch/x86/tpm.c
```diff
@@ -978,11 +971,150 @@ void tpm_hash_extend(unsigned loc, unsigned pcr, uint8_t *buf, unsigned size,
...
+    if ( policy->revision != SLR_POLICY_REVISION )
+        TPM_PANIC("DRTM policy in SLRT is of unsupported revision: %#04x!\n",
+                  slrt->revision);
+    if ( policy->nr_entries == 0 )
+        TPM_PANIC("DRTM policy in SLRT is empty!\n");
+
+    policy_entry = (struct slr_policy_entry *)
+        ((uint8_t *)policy + sizeof(*policy));
+    if ( policy_entry->entity_type != SLR_ET_SLRT )
+        TPM_PANIC("First entry of DRTM policy in SLRT is empty!\n");
```
krystian-hebel: Message is inconsistent with check, maybe something like:

```suggestion
        TPM_PANIC("First entry of DRTM policy in SLRT has wrong type: %#04x!\n",
                  policy_entry->entity_type);
```

xen/include/xen/slr_table.h
```diff
@@ -0,0 +1,254 @@
...
+ */
+
+#ifndef _SLR_TABLE_H
+#define _SLR_TABLE_H
+
+#define UEFI_SLR_TABLE_GUID \
+    { 0x877a9b2a, 0x0385, 0x45d1, { 0xa0, 0x34, 0x9d, 0xac, 0x9c, 0x9e, 0x56, 0x5f }}
+
+/* SLR table header values */
+#define SLR_TABLE_MAGIC		0x4452544d
```
krystian-hebel: https://github.com/TrenchBoot/xen/blob/master/CODING_STYLE#L36 - use spaces instead of tabs everywhere in this file.

xen/include/xen/slr_table.h
```diff
@@ -0,0 +1,254 @@
+/*
+ *  SPDX-License-Identifier: GPL-3.0
```
krystian-hebel: Xen coding style has requirements for [single-line SPDX header](https://github.com/TrenchBoot/xen/blob/master/CODING_STYLE#L20), but there are many files that don't follow it, or even don't have SPDX tags at all, so I'm not sure what to do about it. Perhaps it would be safest to move this to separate comment, but in any case we will have to re-check for proper licenses and copyrights before upstreaming, so we can postpone it until then and check all our changes at once.

xen/arch/x86/include/asm/intel_txt.h
```diff
@@ -341,22 +325,51 @@ static inline int is_in_pmr(struct txt_os_sinit_data *os_sinit, uint64_t base,
 	return 0;
 }
 
+static inline void find_evt_log(void **evt_log, uint32_t *evt_log_size)
```
krystian-hebel, line 344: Please add a comment that `evt_log` is physical address and it is up to the caller to map it to virtual, if needed.

xen/arch/x86/tpm.c
```diff
@@ -978,11 +971,150 @@ void tpm_hash_extend(unsigned loc, unsigned pcr, uint8_t *buf, unsigned size,
...
+    policy_entry = (struct slr_policy_entry *)
+        ((uint8_t *)policy + sizeof(*policy));
+
+    for (i = 1; i < policy->nr_entries; i++) {
+        uint64_t start = policy_entry[i].entity;
+        uint64_t size = policy_entry[i].size;
+        uint32_t event_type = TXT_EVTYPE_UNKNOWN;
+
+        switch (policy_entry[i].entity_type) {
+        case SLR_ET_SLRT:
```
krystian-hebel, line 1131: This needs special treatment, described in Appendix A of Secure Launch Specification. SLRT contains (among other things) addresses that may change between boots, these are not measured.

xen/arch/x86/tpm.c
SergiiDmytruk: I might have taken this note of yours too literally:

> since SLRT will hold the address of MB2 tags, it must be measured and partially parsed very early, possibly in assembly

xen/arch/x86/tpm.c
SergiiDmytruk: That appendix is actually about DRTM, which is one of multiple SLRT entries that contains pointers (there are DCE, DL handler, bootloader context and probably other fields).  Hashing of SLRT then also needs to be done by extending hash with each entry in turn taking pointer values into account and "recursively" processing DRTM.

As mentioned in https://github.com/TrenchBoot/grub/pull/14#discussion_r1388220696, probably need some consistent policy on such things.  Left code as is for now.

xen/arch/x86/include/asm/intel_txt.h
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/5ed262bc0df7696f824b5a547339a804fa9e5a52..39a1478db53029b133c75d4f15a0f5774ad92f8a#diff-dd4611d055ace3b3c87ffe90aaf480b0d4a0e4098adb6de5026e3144a4821b09R328-R329

xen/include/xen/slr_table.h
SergiiDmytruk: Might as well do it now: https://github.com/TrenchBoot/xen/compare/5ed262bc0df7696f824b5a547339a804fa9e5a52..39a1478db53029b133c75d4f15a0f5774ad92f8a#diff-609762c7fad93e82743d840bbbfd586771646f31761b1b8bd54bd4384a591dfbR1-R2

xen/include/xen/slr_table.h
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/5ed262bc0df7696f824b5a547339a804fa9e5a52..39a1478db53029b133c75d4f15a0f5774ad92f8a#diff-609762c7fad93e82743d840bbbfd586771646f31761b1b8bd54bd4384a591dfbL15-R67

xen/arch/x86/tpm.c
SergiiDmytruk: Thanks: https://github.com/TrenchBoot/xen/compare/5ed262bc0df7696f824b5a547339a804fa9e5a52..39a1478db53029b133c75d4f15a0f5774ad92f8a#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1037-R1038

xen/arch/x86/tpm.c
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/5ed262bc0df7696f824b5a547339a804fa9e5a52..39a1478db53029b133c75d4f15a0f5774ad92f8a#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1036

xen/arch/x86/tpm.c
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/5ed262bc0df7696f824b5a547339a804fa9e5a52..39a1478db53029b133c75d4f15a0f5774ad92f8a#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeL27-R1039

krystian-hebel: This is a chicken and egg problem, especially since information in those structures partially overlaps. We could resign from even passing MBI address directly and obtain it from SLRT instead, but I think that as long as we don't have a way of executing code that wasn't measured (i.e. modules) before SLRT is parsed it can be postponed.

Sorry for changing my mind on the matter and not communicating it earlier.

xen/arch/x86/tpm.c
krystian-hebel: Fair point. Perhaps we should organize a (public? not sure how many people would be interested in attending) meeting for discussing everything that specification lacks, but there are still many issues being discovered so maybe now would be too soon.

xen/arch/x86/tpm.c
```diff
@@ -985,4 +978,106 @@ void __stdcall tpm_extend_mbi(uint32_t *mbi)
...
+        case SLR_ET_UEFI_MEMMAP:
+            event_type = TXT_EVTYPE_UEFI_MMAP;
+            break;
+        case SLR_ET_RAMDISK:
+            event_type = TXT_EVTYPE_INITRD;
+            break;
+        case SLR_ET_TXT_OS2MLE:
+            event_type = TXT_EVTYPE_OS2MLE;
+            break;
+        case SLR_ET_MULTIBOOT_MODULE:
```
krystian-hebel, line 1140: You should drop https://github.com/TrenchBoot/xen/commit/7962cb8b862beabd9de2c78f9dae5b02530e7f07, otherwise kernel and initramfs would be measured twice.

It would be good to check if modules parsed by Xen are the same as here (location and size). At the point where `tpm_take_measurements()` is called modules are not relocated yet, and their addresses are still in MB2 format, so it should be relatively easy to do. With that done it should be safe to drop "Excessive multiboot modules for slaunch - limiting to 2" logic from setup.c, assuming no unmeasured modules are left in MBI.

SergiiDmytruk: > You should drop [7962cb8](https://github.com/TrenchBoot/xen/commit/7962cb8b862beabd9de2c78f9dae5b02530e7f07), otherwise kernel and initramfs would be measured twice.

Reverted in cd2f2962e88b206f8d289424d2f1b1335f3a7d04 and 2376e6160348fa439e77918e9f976671ad63e3b9.

> It would be good to check if modules parsed by Xen are the same as here (location and size).

https://github.com/TrenchBoot/xen/compare/39a1478db53029b133c75d4f15a0f5774ad92f8a..bab60f90c52351add03aec69881f03a9c30cda6a#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24ae

SergiiDmytruk:
CI seems to fail because of recent changes in https://github.com/QubesOS/qubes-vmm-xen.  I think `smp_cleanup_upstreaming` was recently rebased to take this into account.

krystian-hebel:
`smp_cleanup_upstreaming` was rebased just for upstreaming changes not directly related to AEM. I think that the conflict is caused by https://github.com/QubesOS/qubes-vmm-xen/blob/main/0512-xsa446.patch, I also had to do some manual work around it during rebasing.

I'm not sure what would be the best approach to deal with it. We could:
- include QubesOS-specific patches in `aem`, but this would make every bump on Qubes side much harder for us, requiring another rebase each time it happens,
- make CI do a rebase on-the-fly, resolving conflicts in the process (patches to patch patches on top of patches...),
- wait for https://github.com/QubesOS/qubes-vmm-xen to switch to Xen 4.18 (IIUC it was released last Friday so it may be soon, maybe @marmarek would know the expected bump date), which includes XSA fixes.

SergiiDmytruk:
> * include QubesOS-specific patches in `aem`, but this would make every bump on Qubes side much harder for us, requiring another rebase each time it happens,

Could apply only those patches which cause conflicts if that's possible.

> * make CI do a rebase on-the-fly, resolving conflicts in the process (patches to patch patches on top of patches...),

I think `format-patch` + `am` essentially amounts to what `git rebase` would do (`rebase` generates its temporary patches a bit differently, but I don't think that would help), so it might result in a similar failure when conflicts can't be resolved automatically.

xen/arch/x86/tpm.c
```diff
@@ -978,4 +978,136 @@ void __stdcall tpm_extend_mbi(uint32_t *mbi)
...
+}
+
+static void verify_drtm_mbi_consistency(struct slr_entry_policy *policy,
+                                        struct slr_policy_entry *policy_entry,
+                                        multiboot_info_t *mbi)
+{
+    module_t *mods = __va(mbi->mods_addr);
+    uint32_t i;
+
+    for ( i = 0; i < mbi->mods_count; i++ ) {
```
krystian-hebel, line 1069: Can we also check that number of policy entries of type `SLR_ET_MULTIBOOT_MODULE` equals `mbi->mods_count`?

xen/arch/x86/tpm.c
```diff
@@ -985,4 +978,136 @@ void __stdcall tpm_extend_mbi(uint32_t *mbi)
     tpm_hash_extend(DRTM_LOC, DRTM_DATA_PCR, (uint8_t *)mbi, *mbi,
                     TXT_EVTYPE_MBI, NULL, 0);
 }
+#else
+static struct slr_entry_policy *tpm_get_policy(void)
```
krystian-hebel: This function has nothing to do with TPM.
```suggestion
static struct slr_entry_policy *slr_get_policy(void)
```

xen/arch/x86/include/asm/intel_txt.h
```diff
@@ -120,32 +122,16 @@ static inline void txt_reset(uint32_t error)
...
-
 /*
  * Secure Launch defined OS/MLE TXT Heap table
  */
 struct txt_os_mle_data {
 	uint32_t version;
 	uint32_t boot_params_addr;
-	uint64_t saved_misc_enable_msr;
-	struct txt_mtrr_state saved_bsp_mtrrs;
+    uint32_t slrt;
```
krystian-hebel, line 131: Inconsistent indentation, but I've got a feeling that the rest of the file is wrong. Let's leave it in this state, it will have to be changed in existing commit during rebase/squash/upstream.

SergiiDmytruk: Yeah, almost all other lines are indented with tabs.

xen/arch/x86/tpm.c
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/bab60f90c52351add03aec69881f03a9c30cda6a..d820e3dac46047e7e81db6fdc30d72041e9d208e#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeL982-R982

xen/arch/x86/tpm.c
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/bab60f90c52351add03aec69881f03a9c30cda6a..d820e3dac46047e7e81db6fdc30d72041e9d208e#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1028-R1059

SergiiDmytruk:
@krystian-hebel So, to move this and GRUB changes forward, measuring only MBI and measuring SLRT as described in the specification?  If this turns out to be problematic, it's unlikely to be the only thing that wasn't done properly right away, so might be unwise to get stuck on it.  It's not even large amount of code either way.

krystian-hebel:
I'd say MBI, modules and SLRT as in `sl_extend_slrt` in [patches sent to Linux](https://lkml.org/lkml/2023/11/10/728):

```
/*
* In revision one of the SLRT, the only table that needs to be
* measured is the Intel info table. Everything else is meta-data,
* addresses and sizes. Note the size of what to measure is not set.
* The flag SLR_POLICY_IMPLICIT_SIZE leaves it to the measuring code
* to sort out.
*/
```

[Patch 05/13 of that series](https://lkml.org/lkml/2023/11/10/725) also has different definitions for `TXT_EVTYPE_*` constants, we probably should align with those.

SergiiDmytruk:
> [Patch 05/13 of that series](https://lkml.org/lkml/2023/11/10/725) also has different definitions for `TXT_EVTYPE_*` constants, we probably should align with those.

There is this  comment in `intel_txt.h`:
```c
/* TXT-defined use 0x4xx, TrenchBoot in Linux uses 0x5xx, use 0x6xx here. */
```

xen/arch/x86/tpm.c
SergiiDmytruk: New way of handling SLRT as per https://github.com/TrenchBoot/xen/pull/7#issuecomment-1821023039: https://github.com/TrenchBoot/xen/compare/d820e3dac46047e7e81db6fdc30d72041e9d208e..6ec1f7b7ae9af57ca6abbb7729255aaf14fbcfa7#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1113-R1127

SergiiDmytruk:
Also started handling DRTM entry's flags:
1. https://github.com/TrenchBoot/xen/compare/d820e3dac46047e7e81db6fdc30d72041e9d208e..6ec1f7b7ae9af57ca6abbb7729255aaf14fbcfa7#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1040-R1042
2. https://github.com/TrenchBoot/xen/compare/d820e3dac46047e7e81db6fdc30d72041e9d208e..6ec1f7b7ae9af57ca6abbb7729255aaf14fbcfa7#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1059-R1061
3. https://github.com/TrenchBoot/xen/compare/d820e3dac46047e7e81db6fdc30d72041e9d208e..6ec1f7b7ae9af57ca6abbb7729255aaf14fbcfa7#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1099-R1106

xen/arch/x86/tpm.c
krystian-hebel: Could you reverse the message in `panic()`? Policy in SLRT is the source of truth, and MBI should match it, not the other way around. Current message suggests that policy is wrong in case of mismatch.

krystian-hebel:
> Also started handling DRTM entry's flags: (...)

I think that check for `SLR_POLICY_FLAG_MEASURED` for modules (1. and 2.) should be removed. We don't use it in any way right now, but if we ever will (e.g. if some low-level module is consumed very early, earlier than `verify_drtm_mbi_consistency()` is called) it would make sense to measure it earlier and set mentioned flag without actually removing the module from MBI. We would still need to check whether a) module exists in MBI and b) has the same address and size.

As for 3rd link, the code should test if entries that are already measured are first entries in the table. Order of measurements matters, it should reflect the order of policy entries. So if an entry that is not yet measured is found, no further entry may have `SLR_POLICY_FLAG_MEASURED` set. Ideally, error message should mention that order of modules in grub.cfg should match that of policy, because IIUC this is how GRUB orders modules in MBI.

> There is this comment in intel_txt.h: (...)

I know, I wrote it :slightly_smiling_face: This was before any specification was even planned and design decisions were still being made, so I wanted to avoid any potential conflicts. Now those types are somewhat defined, and most entries use the same type, which is different to what I planned back then. Having fewer types will make it easier to add support for other payloads/OSes/hypervisors without having to add new types, propagate them and potentially resolve any conflicts.

xen/arch/x86/tpm.c
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/6ec1f7b7ae9af57ca6abbb7729255aaf14fbcfa7..c8f93f26c569ab37177393c2ceedb0cd012da4d1#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeL1066-R1063

SergiiDmytruk:
Updated constants and flags handling in https://github.com/TrenchBoot/xen/compare/6ec1f7b7ae9af57ca6abbb7729255aaf14fbcfa7..c8f93f26c569ab37177393c2ceedb0cd012da4d1#

krystian-hebel:
I've tried booting this + GRUB on hardware and it seems to reboot somewhere in Xen, before anything is printed on the screen. I don't have a way of connecting to serial at the moment, so I can't debug any further.

SergiiDmytruk:
What about GRUB changes + first two commits of this PR?  That should work if bug doesn't show up without the last commit.

If it works and deployment of new code isn't too hard, it's possible to bisect changes by adding `return` somewhere in `tpm_take_measurements()` to narrow it down a bit (everything should work if that code doesn't run at all).  Could be that GRUB provides incorrect data and Xen code is fine, knowing approximately how far things go should make code inspection easier either way.

krystian-hebel:
> What about GRUB changes + first two commits of this PR?

Same. I ended up adding infinite loop as the first Xen instruction and it still reboots, so the issue is most likely in GRUB. Unfortunately, it seems that either BIOS clears TXT.ERRORCODE on boot or reboot happens before SINIT. In any case, I'm moving the review to GRUB PR.

xen/arch/x86/intel_txt.c
```diff
@@ -92,13 +94,14 @@ void __init protect_txt_mem_regions(void)
...
         /* TXT TPM Event Log */
         map_l2(txt_heap_base, txt_heap_size);
-        os_mle = txt_os_mle_data_start(__va(txt_heap_base));
-
-        if ( os_mle->evtlog_addr != 0 ) {
-            printk("SLAUNCH: reserving event log (%#lx - %#lx)\n", os_mle->evtlog_addr,
-                   os_mle->evtlog_addr + os_mle->evtlog_size);
-            e820_change_range_type(&e820_raw, os_mle->evtlog_addr,
-                                   os_mle->evtlog_addr + os_mle->evtlog_size,
+        find_evt_log(&evt_log_addr, &evt_log_size);
```
krystian-hebel, line 72: This call fails because SLRT isn't mapped in non-early environment. While we could map it (and reserve to protect against overwrites), maybe we should locate it in TXT heap, in OS-to-MLE section instead? There is plenty of space available, and having SLRT on the TXT heap would make it much easier to handle because it is already protected (both from DMA and in memory map), mapped and doesn't cause additional fragmentation of physical address space.

SergiiDmytruk: How much space is there typically?  Current GRUB code allocates 4KiB of 4KiB-aligned memory in the range 0x1000000 to 0xFFFFEFFF.  Can we guarantee such alignment?  I guess we can always fail if not.

krystian-hebel: On platform I have at hand heap has 2M - 128K, out of it 0x43A bytes are currently used, including OS-MLE and all *DataSize fields.

It doesn't have to be aligned to a page, Secure Launch specification requires only 4-byte alignment. TXT specification says that `For proper data alignment on 64-bit processor architectures all table sizes must be a multiple of 8 bytes` when it talks about data on heap, but first entry (produced by BIOS ACM) already violates this, so we can't rely on any OS-MLE alignment.

SergiiDmytruk: https://github.com/TrenchBoot/grub/compare/03579a740622f9fb5002f666b8bd4a93e40696a9..e1bdaf2ef7638f56b9c059639b91a9c36596bfc1

xen/arch/x86/include/asm/intel_txt.h
```diff
@@ -341,22 +327,51 @@ static inline int is_in_pmr(struct txt_os_sinit_data *os_sinit, uint64_t base,
...
 
+/* evt_log is a physical address and the caller must map it to virtual, if
+ * needed. */
+static inline void find_evt_log(void **evt_log, uint32_t *evt_log_size)
+{
+    struct txt_os_mle_data *os_mle;
+    struct slr_table *slrt;
+    struct slr_entry_log_info *log_info;
+
+    os_mle = txt_os_mle_data_start(_txt(read_txt_reg(TXTCR_HEAP_BASE)));
```
krystian-hebel, line 350: `read_txt_reg()` requires TXT register space and `txt_os_mle_data_start()` - TXT heap to be mapped.

xen/arch/x86/intel_txt.c
```diff
@@ -82,7 +83,8 @@ void __init protect_txt_mem_regions(void)
```
krystian-hebel: Applies to lines above but GH won't let me mark them:
- `map_l2()` should use `PRIV` instead of `PUB`, as this is what `{read,write}_txt_reg()` uses.
- I copied `map_l2()` to another file because I'm lazy, but maybe it should be exposed or made an inline in some header file instead.
- `unmap_l2()` probably can be dropped altogether. A lot of further code accesses TXT registers and heap, and Xen will rebuild memory maps soon after this function is called, effectively dropping all existing mappings.

xen/arch/x86/intel_txt.c
```diff
@@ -92,13 +94,14 @@ void __init protect_txt_mem_regions(void)
...
+        find_evt_log(&evt_log_addr, &evt_log_size);
+
+        if ( evt_log_addr != 0 ) {
+            printk("SLAUNCH: reserving event log (%#lx - %#lx)\n",
+                   (uint64_t)evt_log_addr,
+                   (uint64_t)evt_log_addr + evt_log_size);
+            e820_change_range_type(&e820_raw, (uint64_t)evt_log_addr,
+                                   (uint64_t)evt_log_addr + evt_log_size,
                                    E820_RAM, E820_RESERVED);
         }
```
krystian-hebel: `unmap_l2()` from next line may be dropped. In that case, `map_l2()`/`unmap_l2()` pair in `txt_restore_mtrrs()` should also be removed.

xen/arch/x86/setup.c
```diff
@@ -1030,11 +1030,15 @@ void __init noreturn __start_xen(unsigned long mbi_p)
     }
 
     if ( sl_status )
+    {
+        tpm_take_measurements(mbi);
```
krystian-hebel: This is too early for this call. For testing I moved it right after `protect_txt_mem_regions()`, which left TXT regs and heap mapped and it kind of worked, but booting took tens of minutes due to hashing modules located in UC memory. After moving it after `init_e820()` which restores MTRRs it worked nicely, although we should measure at least SLRT entry before using MTRRs saved there. Most likely `tpm_take_measurements()` needs to be split further.

xen/arch/x86/tpm.c
```diff
@@ -933,6 +925,7 @@ void tpm_hash_extend(unsigned loc, unsigned pcr, uint8_t *buf, unsigned size,
     uint32_t evt_log_size;
 
     find_evt_log(&evt_log_addr, &evt_log_size);
+    evt_log_addr = __va(evt_log_addr);
```
krystian-hebel, line 928: `__va()` requires that event log is mapped. Maybe map it once in `protect_txt_mem_regions()` and keep it mapped?

xen/arch/x86/tpm.c
```diff
@@ -983,6 +976,181 @@ void __stdcall tpm_extend_mbi(uint32_t *mbi)
...
-                    TXT_EVTYPE_MBI, NULL, 0);
+                    TXT_EVTYPE_SLAUNCH, NULL, 0);
+}
+#else
+static struct slr_table *slr_get_table(void)
+{
+    struct txt_os_mle_data *os_mle;
+    struct slr_table *slrt;
+
+    os_mle = txt_os_mle_data_start(__va(read_txt_reg(TXTCR_HEAP_BASE)));
```
krystian-hebel, line 973: Needs TXT regs and heap mapped.

xen/arch/x86/tpm.c
```diff
@@ -983,6 +976,181 @@ void __stdcall tpm_extend_mbi(uint32_t *mbi)
...
+
+    slrt = __va(os_mle->slrt);
```
krystian-hebel, line 976: `__va()` needs SLRT to be mapped, in current form this is handled by mapping TXT heap, but maybe it wouldn't hurt to map it separately, just in case.

xen/arch/x86/tpm.c
```diff
@@ -983,6 +976,181 @@ void __stdcall tpm_extend_mbi(uint32_t *mbi)
...
+        case SLR_ET_MULTIBOOT_INFO:
+            panic("Duplicated MBI entry in DRTM of Secure Launch at %d\n", i);
+
+        case SLR_ET_SLRT:
+            if ( slrt->revision == 1 ) {
+                /*
+                 * In revision one of the SLRT, only Intel info table is
+                 * measured.  It's presence is checked by slr_get_policy().
+                 */
+                start = (uint64_t)slr_next_entry_by_tag(__va(start), NULL,
```
krystian-hebel: `start` is made virtual here, but other entry types have physical `start`.

We should probably check if this is the same as physical address of SLRT read from OS-MLE (`__pa(start) == __pa(slrt)` or something similar). The idea of DRTM is to treat everything as untrusted, including bootloader.

xen/arch/x86/tpm.c
```diff
@@ -983,6 +976,181 @@ void __stdcall tpm_extend_mbi(uint32_t *mbi)
...
+
+        case SLR_ET_UNUSED:
+            /* Skip this entry. */
+            continue;
+        }
+
+        if ( size == 0 )
+            panic("Can't measure DRTM policy entry at %d of size zero!\n", i);
+
+        map_pages_to_xen(start, maddr_to_mfn(start), PFN_UP(size),
```
krystian-hebel: At the point it is called, `map_l2(start, size)` must be used. Also, this breaks for SLRT because `start` used there is already virtual.

xen/arch/x86/tpm.c
```diff
@@ -983,6 +976,181 @@ void __stdcall tpm_extend_mbi(uint32_t *mbi)
...
+                         __PAGE_HYPERVISOR_RO);
+        tpm_hash_extend(DRTM_LOC, policy_entry[i].pcr, (uint8_t *)start, size,
```
krystian-hebel: Except for SLRT:
```suggestion
        tpm_hash_extend(DRTM_LOC, policy_entry[i].pcr, (uint8_t *)__va(start), size,
```

xen/arch/x86/tpm.c
```diff
@@ -983,6 +976,181 @@ void __stdcall tpm_extend_mbi(uint32_t *mbi)
...
+                        TXT_EVTYPE_SLAUNCH, NULL, 0);
+        destroy_xen_mappings(start, start + PFN_UP(size) * PAGE_SIZE);
```
krystian-hebel: This can be removed, we don't want to remove mapping to TXT heap accidentally and there should be no consequences of leaving it mapped.

xen/arch/x86/tpm.c
```diff
@@ -983,6 +976,181 @@ void __stdcall tpm_extend_mbi(uint32_t *mbi)
...
+    }
+
+    verify_drtm_mbi_consistency(policy, policy_entry, mbi);
+
+    /*
+     * All SLR_POLICY_FLAG_MEASURED must be at the start of the list.
+     * MBI was measured in tpm_extend_mbi().
+     */
+    for ( i = 1; i < policy->nr_entries; i++ ) {
+        if ( !(policy_entry[i].flags & SLR_POLICY_FLAG_MEASURED) )
```
krystian-hebel: Do we even set this bit? If not, we probably should.

xen/arch/x86/setup.c
```diff
@@ -1030,11 +1030,15 @@ void __init noreturn __start_xen(unsigned long mbi_p)
...
+
         if ( mbi->mods_count > 2 )
```
krystian-hebel: This can now be removed, along with checks before `xsm_multiboot_init()` and `microcode_grab_module()` further in this file. Both are passed as Multiboot2 modules so they are now measured as part of DRTM policy. I think all those were added in one commit so it should be possible to just revert it.

krystian-hebel (commented):
These should be all of the changes required to boot successfully.

xen/arch/x86/tpm.c
krystian-hebel: For modules specifically there is `bootstrap_map()` which should do the same as `map_l2()` and it already is externally visible. Of the rest of entry types, only `SLRT` and `TXT_OS2MLE` are used and still left to be measured, but those are mapped as part of TXT heap.

xen/arch/x86/tpm.c
```diff
@@ -983,6 +976,181 @@ void __stdcall tpm_extend_mbi(uint32_t *mbi)
...
+            continue;
+        }
+
+        if ( size == 0 )
+            panic("Can't measure DRTM policy entry at %d of size zero!\n", i);
+
+        map_pages_to_xen(start, maddr_to_mfn(start), PFN_UP(size),
+                         __PAGE_HYPERVISOR_RO);
+        tpm_hash_extend(DRTM_LOC, policy_entry[i].pcr, (uint8_t *)start, size,
+                        TXT_EVTYPE_SLAUNCH, NULL, 0);
```
krystian-hebel:
```suggestion
                        TXT_EVTYPE_SLAUNCH, policy_entry[i].evt_info, strlen(policy_entry[i].evt_info));
```

Maybe there is some more secure `strlen()` version, haven't checked.

xen/arch/x86/include/asm/intel_txt.h
SergiiDmytruk: I guess if we map all used regions early enough, this is fine as is.

xen/arch/x86/intel_txt.c
SergiiDmytruk: See https://github.com/TrenchBoot/xen/pull/7/commits/bdfad3c6cb2fad93de1c998d6c072f39a1c039dc

xen/arch/x86/intel_txt.c
SergiiDmytruk: bdfad3c6cb2fad93de1c998d6c072f39a1c039dc should be addressing this.

xen/arch/x86/setup.c
SergiiDmytruk: Moved things around, added separate functions to map TXT regions and to measure SLRT.

https://github.com/TrenchBoot/xen/compare/c8f93f26c569ab37177393c2ceedb0cd012da4d1..a0e417e572aae4976aaa74d4e3cc36cf2ebaa25e#diff-961c78e908063b25617eab27634027bfd5ba6bd6d9253c53a5ac322b94a7c791

xen/arch/x86/setup.c
SergiiDmytruk: Done in https://github.com/TrenchBoot/xen/compare/c8f93f26c569ab37177393c2ceedb0cd012da4d1..a0e417e572aae4976aaa74d4e3cc36cf2ebaa25e#diff-961c78e908063b25617eab27634027bfd5ba6bd6d9253c53a5ac322b94a7c791L1032-L1042

xen/arch/x86/tpm.c
SergiiDmytruk: bdfad3c6cb2fad93de1c998d6c072f39a1c039dc makes maps permanent.

xen/arch/x86/tpm.c
SergiiDmytruk: I believe with https://github.com/TrenchBoot/xen/commit/bdfad3c6cb2fad93de1c998d6c072f39a1c039dc it should work as is.

xen/arch/x86/tpm.c
SergiiDmytruk: Done in https://github.com/TrenchBoot/xen/compare/c8f93f26c569ab37177393c2ceedb0cd012da4d1..a0e417e572aae4976aaa74d4e3cc36cf2ebaa25e#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR989-R1005

xen/arch/x86/tpm.c
SergiiDmytruk: Added:
1. https://github.com/TrenchBoot/xen/compare/c8f93f26c569ab37177393c2ceedb0cd012da4d1..a0e417e572aae4976aaa74d4e3cc36cf2ebaa25e#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1114
2. https://github.com/TrenchBoot/xen/compare/c8f93f26c569ab37177393c2ceedb0cd012da4d1..a0e417e572aae4976aaa74d4e3cc36cf2ebaa25e#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1124
3. https://github.com/TrenchBoot/xen/compare/c8f93f26c569ab37177393c2ceedb0cd012da4d1..a0e417e572aae4976aaa74d4e3cc36cf2ebaa25e#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1168

xen/arch/x86/tpm.c
SergiiDmytruk: Put the check in https://github.com/TrenchBoot/xen/compare/c8f93f26c569ab37177393c2ceedb0cd012da4d1..cc6e675973ecc45d3e99d0c9b1301f16d3912476#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1123-R1125

xen/arch/x86/tpm.c
SergiiDmytruk: Mapping everything with `map_l2()` for consistency: https://github.com/TrenchBoot/xen/compare/c8f93f26c569ab37177393c2ceedb0cd012da4d1..cc6e675973ecc45d3e99d0c9b1301f16d3912476#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1165

Using `bootstrap_map()` would also need looking up a module by its address.

xen/arch/x86/tpm.c
SergiiDmytruk: In https://github.com/TrenchBoot/xen/compare/c8f93f26c569ab37177393c2ceedb0cd012da4d1..cc6e675973ecc45d3e99d0c9b1301f16d3912476#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1166

xen/arch/x86/tpm.c
SergiiDmytruk: Dropped in https://github.com/TrenchBoot/xen/compare/c8f93f26c569ab37177393c2ceedb0cd012da4d1..cc6e675973ecc45d3e99d0c9b1301f16d3912476#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeL1153

xen/arch/x86/tpm.c
SergiiDmytruk: See https://github.com/TrenchBoot/xen/compare/c8f93f26c569ab37177393c2ceedb0cd012da4d1..cc6e675973ecc45d3e99d0c9b1301f16d3912476#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeR1167-R1169

This however doesn't add description for `MBI` and `SLRT`.

SergiiDmytruk (commented):
There were lots of changes in various places, but hopefully they agree on each other.

I think `verify_drtm_mbi_consistency()` now needs to be called separately earlier (where `tpm_take_measurements()` was called before), but other than that changes should address the review.  Will do this change tomorrow.

xen/arch/x86/setup.c
krystian-hebel: What about removing checks before mentioned functions? They have `TODO` in comment and `if (sl_status)`, those can be now dropped and functions can be called unconditionally.

xen/arch/x86/tpm.c
krystian-hebel: Let's leave those descriptions as possible future improvements. I understand that this implies depending on SLRT before it is measured, but in this case it isn't critical as it doesn't change the execution flow and TPM event log is purely informational.

xen/arch/x86/tpm.c
```diff
@@ -933,6 +925,7 @@ void tpm_hash_extend(unsigned loc, unsigned pcr, uint8_t *buf, unsigned size,
     uint32_t evt_log_size;
 
     find_evt_log(&evt_log_addr, &evt_log_size);
+    evt_log_addr = __va(evt_log_addr);
 
 #ifndef __EARLY_TPM__
```
krystian-hebel: Mapping protected by this should be dropped, event log is already mapped and this results in exception because region is not aligned properly. Same goes for unmapping at the end of the function.

SergiiDmytruk: See https://github.com/TrenchBoot/xen/compare/cc6e675973ecc45d3e99d0c9b1301f16d3912476..f74bb7a7e506d9beb2e8267a9d3b4572b968b62d#diff-5861d423276bd98c31fb2a501569110297cbe6488842e0359435a3af795c24aeL930-L971

SergiiDmytruk:
> I think `verify_drtm_mbi_consistency()` now needs to be called separately earlier (where `tpm_take_measurements()` was called before), but other than that changes should address the review. Will do this change tomorrow.

That turned out to be unnecessary, modules are updated after DRTM is processed.  I however turned `tpm_take_measurements()` into `check_drtm_policy()` and collected all checks related to DRTM entries there (from `slr_get_policy()` and `tpm_process_drtm_policy()`) as they were too far apart.  See https://github.com/TrenchBoot/xen/compare/cc6e675973ecc45d3e99d0c9b1301f16d3912476..f74bb7a7e506d9beb2e8267a9d3b4572b968b62d

xen/arch/x86/setup.c
SergiiDmytruk: Thanks, I somehow ignored most of that comment and on top of that also skipped notification about your reply... (Came here from Matrix/Trenchboot and saw it.)

Made it via revert in 3dc5991be74c5fd498e2fb9d7ff2b53289abba9d.

krystian-hebel (approved):

PR #10, opened 2024-03-17
Support AMD SKINIT for Slaunch boot

SergiiDmytruk:
Moved some code into `slaunch.c` and `slaunch.h` to avoid putting AMD-specific things into files named `intel_txt.*` and stop putting too much into `tpm.c`.  Better to review by commits.

Also had to work around limitations of early TPM code for getting SLRT pointer and determining CPU vendor.

### DIFF
--- a/xen/arch/x86/boot/head.S
+++ b/xen/arch/x86/boot/head.S
@@ -522,21 +522,33 @@ __start:
         jmp     trampoline_bios_setup
 
 .Lslaunch_proto:
+        /* Upon reaching here, CPU state mostly matches the one setup by the
+         * bootloader with ESP, ESI and EDX being clobbered above. */
+
         /* Save information that TrenchBoot slaunch was used. */
         movb    $1, sym_esi(slaunch_active)
 
-        /* Push arguments to stack and call txt_early_tests(). */
+        /* Prepare space for output parameter of slaunch_early_tests(), which is
+         * a structure of two uint32_t fields. */
+        sub     $8, %esp
+
+        /* Push arguments to stack and call slaunch_early_tests(). */
+        push    %esp                        /* pointer to output structure */
         push    $sym_offs(__2M_rwdata_end)  /* end of target image */
         push    $sym_offs(_start)           /* target base address */
         push    %esi                        /* load base address */
-        call    txt_early_tests
+        call    slaunch_early_tests
 
-        /*
-         * txt_early_tests() returns MBI address, pass it to tpm_extend_mbi()
-         * and store for later in EBX.
-         */
-        push    %eax
-        movl    %eax,%ebx
+        /* Move outputs of slaunch_early_tests() from stack into registers. */
+        pop     %ebx  /* physical MBI address */
+        pop     %edx  /* physical SLRT address */
+
+        /* Save physical address of SLRT for C code. */
+        mov     %edx, sym_esi(slaunch_slrt)
+
+        /* Push arguments to stack and call tpm_extend_mbi(). */
+        push    %edx  /* SLRT address because early code has no slaunch_slrt */
+        push    %ebx  /* MBI address */
         call    tpm_extend_mbi
 
         /* Move magic number expected by Multiboot 2 to EAX and fall through. */
@@ -864,8 +876,8 @@ reloc:
         .incbin "reloc.bin"
 
         ALIGN
-txt_early_tests:
-        .incbin "txt_early.bin"
+slaunch_early_tests:
+        .incbin "slaunch_early.bin"
 
         ALIGN
 tpm_extend_mbi:
--- a/xen/arch/x86/boot/head.S
+++ b/xen/arch/x86/boot/head.S
@@ -534,6 +534,8 @@ __start:
 
         /* Push arguments to stack and call slaunch_early_tests(). */
         push    %esp                        /* pointer to output structure */
+        push    %ebp                        /* Slaunch parameter on AMD */
+        push    %ebx                        /* Multiboot parameter */
         push    $sym_offs(__2M_rwdata_end)  /* end of target image */
         push    $sym_offs(_start)           /* target base address */
         push    %esi                        /* load base address */
--- a/xen/arch/x86/boot/slaunch_early.c
+++ b/xen/arch/x86/boot/slaunch_early.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2023 3mdeb Sp. z o.o. All rights reserved.
+ * Copyright (c) 2022-2024 3mdeb Sp. z o.o. All rights reserved.
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -23,11 +23,18 @@ asm (
     "    .text                         \n"
     "    .globl _start                 \n"
     "_start:                           \n"
-    "    jmp  txt_early_tests          \n"
+    "    jmp  slaunch_early_tests      \n"
     );
 
 #include "defs.h"
 #include "../include/asm/intel_txt.h"
+#include "../include/asm/slaunch.h"
+
+struct early_tests_results
+{
+    uint32_t mbi_pa;
+    uint32_t slrt_pa;
+} __packed;
 
 static void verify_pmr_ranges(struct txt_os_mle_data *os_mle,
                               struct txt_os_sinit_data *os_sinit,
@@ -104,9 +111,10 @@ static void verify_pmr_ranges(struct txt_os_mle_data *os_mle,
     */
 }
 
-uint32_t __stdcall txt_early_tests(uint32_t load_base_addr,
+void __stdcall slaunch_early_tests(uint32_t load_base_addr,
                                    uint32_t tgt_base_addr,
-                                   uint32_t tgt_end_addr)
+                                   uint32_t tgt_end_addr,
+                                   struct early_tests_results *result)
 {
     void *txt_heap;
     struct txt_os_mle_data *os_mle;
@@ -127,5 +135,6 @@ uint32_t __stdcall txt_early_tests(uint32_t load_base_addr,
 
     verify_pmr_ranges(os_mle, os_sinit, load_base_addr, tgt_base_addr, size);
 
-    return os_mle->boot_params_addr;
+    result->mbi_pa = os_mle->boot_params_addr;
+    result->slrt_pa = os_mle->slrt;
 }
--- a/xen/arch/x86/boot/slaunch_early.c
+++ b/xen/arch/x86/boot/slaunch_early.c
@@ -29,12 +29,23 @@ asm (
 #include "defs.h"
 #include "../include/asm/intel_txt.h"
 #include "../include/asm/slaunch.h"
+#include "../include/asm/x86-vendors.h"
 
 struct early_tests_results
 {
     uint32_t mbi_pa;
     uint32_t slrt_pa;
 } __packed;
+
+static bool is_intel_cpu(void)
+{
+    /* No boot_cpu_data in early code. */
+    uint32_t eax, ebx, ecx, edx;
+    cpuid(0x00000000, &eax, &ebx, &ecx, &edx);
+    return ebx == X86_VENDOR_INTEL_EBX
+        && ecx == X86_VENDOR_INTEL_ECX
+        && edx == X86_VENDOR_INTEL_EDX;
+}
 
 static void verify_pmr_ranges(struct txt_os_mle_data *os_mle,
                               struct txt_os_sinit_data *os_sinit,
@@ -114,12 +125,31 @@ static void verify_pmr_ranges(struct txt_os_mle_data *os_mle,
 void __stdcall slaunch_early_tests(uint32_t load_base_addr,
                                    uint32_t tgt_base_addr,
                                    uint32_t tgt_end_addr,
+                                   uint32_t multiboot_param,
+                                   uint32_t slaunch_param,
                                    struct early_tests_results *result)
 {
     void *txt_heap;
     struct txt_os_mle_data *os_mle;
     struct txt_os_sinit_data *os_sinit;
     uint32_t size = tgt_end_addr - tgt_base_addr;
+
+    if ( !is_intel_cpu() )
+    {
+        /*
+         * Not an Intel CPU. Currently the only other option is AMD with SKINIT
+         * and secure-kernel-loader.
+         */
+
+        const uint16_t *sl_header = (void *)slaunch_param;
+        /* secure-kernel-loader passes MBI as a parameter for Multiboot
+         * kernel. */
+        result->mbi_pa = multiboot_param;
+        /* The forth 16-bit integer of SKL's header is an offset to
+         * bootloader's data, which is SLRT. */
+        result->slrt_pa = slaunch_param + sl_header[3];
+        return;
+    }
 
     /* Clear the TXT error registers for a clean start of day */
     write_txt_reg(TXTCR_ERRORCODE, 0);
--- a/xen/arch/x86/include/asm/slaunch.h
+++ b/xen/arch/x86/include/asm/slaunch.h
@@ -11,6 +11,8 @@
 /*
  * Secure Launch event log entry types. The TXT specification defines the
  * base event value as 0x400 for DRTM values.
+ *
+ * Using the same values for AMD SKINIT.
  */
 #define TXT_EVTYPE_BASE            0x400
 #define DLE_EVTYPE_SLAUNCH         (TXT_EVTYPE_BASE + 0x102)
--- a/xen/arch/x86/include/asm/slaunch.h
+++ b/xen/arch/x86/include/asm/slaunch.h
@@ -1,0 +1,65 @@
+#ifndef _ASM_X86_SLAUNCH_H_
+#define _ASM_X86_SLAUNCH_H_
+
+#include <xen/types.h>
+#include <xen/multiboot.h>
+#include <xen/slr_table.h>
+
+#define DRTM_LOC                   2
+#define DRTM_CODE_PCR              17
+#define DRTM_DATA_PCR              18
+
+/*
+ * Secure Launch event log entry types. The TXT specification defines the
+ * base event value as 0x400 for DRTM values.
+ */
+#define TXT_EVTYPE_BASE            0x400
+#define DLE_EVTYPE_SLAUNCH         (TXT_EVTYPE_BASE + 0x102)
+#define DLE_EVTYPE_SLAUNCH_START   (TXT_EVTYPE_BASE + 0x103)
+#define DLE_EVTYPE_SLAUNCH_END     (TXT_EVTYPE_BASE + 0x104)
+
+extern bool slaunch_active;
+
+/* evt_log is a physical address and the caller must map it to virtual, if
+ * needed. */
+static inline void find_evt_log(struct slr_table *slrt, void **evt_log,
+                                uint32_t *evt_log_size)
+{
+    struct slr_entry_log_info *log_info =
+        (void *)slr_next_entry_by_tag(slrt, NULL, SLR_ENTRY_LOG_INFO);
+
+    if ( log_info != NULL ) {
+        *evt_log = _p(log_info->addr);
+        *evt_log_size = log_info->size;
+    } else {
+        *evt_log = NULL;
+        *evt_log_size = 0;
+    }
+}
+
+/*
+ * This helper function is used to map memory using L2 page tables by aligning
+ * mapped regions to 2MB. This way page allocator (which at this point isn't
+ * yet initialized) isn't needed for creating new L1 mappings. The function
+ * also checks and skips memory already mapped by the prebuilt tables.
+ *
+ * There is no unmap_l2() because the function is meant to be used for code that
+ * accesses TXT registers and TXT heap soon after which Xen rebuilds memory
+ * maps, effectively dropping all existing mappings.
+ */
+extern int map_l2(unsigned long paddr, unsigned long size);
+
+void tpm_hash_extend(unsigned loc, unsigned pcr, uint8_t *buf, unsigned size,
+                     uint32_t type, uint8_t *log_data, unsigned log_data_size);
+
+/* Measures essential parts of SLR table before making use of them. */
+void tpm_measure_slrt(void);
+
+/* Takes measurements of DRTM policy entries except for MBI and SLRT which
+ * should have been measured by the time this is called. Also performs sanity
+ * checks of the policy and panics on failure. In particular, the function
+ * verifies that DRTM is consistent with MultibootInfo (MBI) (the MBI address
+ * is assumed to be virtual). */
+void tpm_process_drtm_policy(const multiboot_info_t *mbi);
+
+#endif /* _ASM_X86_SLAUNCH_H_ */
--- a/xen/arch/x86/include/asm/slaunch.h
+++ b/xen/arch/x86/include/asm/slaunch.h
@@ -18,6 +18,7 @@
 #define DLE_EVTYPE_SLAUNCH_END     (TXT_EVTYPE_BASE + 0x104)
 
 extern bool slaunch_active;
+extern uint32_t slaunch_slrt; /* physical address */
 
 /* evt_log is a physical address and the caller must map it to virtual, if
  * needed. */
--- a/xen/arch/x86/include/asm/slaunch.h
+++ b/xen/arch/x86/include/asm/slaunch.h
@@ -2,7 +2,6 @@
 #define _ASM_X86_SLAUNCH_H_
 
 #include <xen/types.h>
-#include <xen/multiboot.h>
 #include <xen/slr_table.h>
 
 #define DRTM_LOC                   2
@@ -48,18 +47,5 @@ static inline void find_evt_log(struct slr_table *slrt, void **evt_log,
  * maps, effectively dropping all existing mappings.
  */
 extern int map_l2(unsigned long paddr, unsigned long size);
-
-void tpm_hash_extend(unsigned loc, unsigned pcr, uint8_t *buf, unsigned size,
-                     uint32_t type, uint8_t *log_data, unsigned log_data_size);
-
-/* Measures essential parts of SLR table before making use of them. */
-void tpm_measure_slrt(void);
-
-/* Takes measurements of DRTM policy entries except for MBI and SLRT which
- * should have been measured by the time this is called. Also performs sanity
- * checks of the policy and panics on failure. In particular, the function
- * verifies that DRTM is consistent with MultibootInfo (MBI) (the MBI address
- * is assumed to be virtual). */
-void tpm_process_drtm_policy(const multiboot_info_t *mbi);
 
 #endif /* _ASM_X86_SLAUNCH_H_ */
--- a/xen/arch/x86/include/asm/slaunch.h
+++ b/xen/arch/x86/include/asm/slaunch.h
@@ -54,6 +54,10 @@ static inline void find_evt_log(struct slr_table *slrt, void **evt_log,
     }
 }
 
+void map_slaunch_mem_regions(void);
+
+void protect_slaunch_mem_regions(void);
+
 /*
  * This helper function is used to map memory using L2 page tables by aligning
  * mapped regions to 2MB. This way page allocator (which at this point isn't
--- a/xen/arch/x86/include/asm/slaunch.h
+++ b/xen/arch/x86/include/asm/slaunch.h
@@ -17,6 +17,23 @@
 #define DLE_EVTYPE_SLAUNCH_START   (TXT_EVTYPE_BASE + 0x103)
 #define DLE_EVTYPE_SLAUNCH_END     (TXT_EVTYPE_BASE + 0x104)
 
+#ifndef cpuid
+/*
+ * Generic CPUID function
+ * clear %ecx since some cpus (Cyrix MII) do not set or clear %ecx
+ * resulting in stale register contents being returned.
+ *
+ * Copied from processor.h because that header can't be included by early code.
+ */
+#define cpuid(_op,_eax,_ebx,_ecx,_edx)          \
+    asm volatile ( "cpuid"                      \
+          : "=a" (*(int *)(_eax)),              \
+            "=b" (*(int *)(_ebx)),              \
+            "=c" (*(int *)(_ecx)),              \
+            "=d" (*(int *)(_edx))               \
+          : "0" (_op), "2" (0) )
+#endif
+
 extern bool slaunch_active;
 extern uint32_t slaunch_slrt; /* physical address */
 
--- a/xen/arch/x86/intel_txt.c
+++ b/xen/arch/x86/intel_txt.c
@@ -6,6 +6,7 @@
 #include <asm/page.h>
 #include <asm/intel_txt.h>
 #include <asm/slaunch.h>
+#include <asm/tpm.h>
 #include <xen/init.h>
 #include <xen/mm.h>
 #include <xen/slr_table.h>
@@ -18,6 +19,7 @@ void __init map_txt_mem_regions(void)
     uint32_t evt_log_size;
 
     map_l2(TXT_PRIV_CONFIG_REGS_BASE, NR_TXT_CONFIG_PAGES * PAGE_SIZE);
+    map_l2(TPM_TIS_BASE, TPM_TIS_SIZE);
 
     txt_heap_base = read_txt_reg(TXTCR_HEAP_BASE);
     BUG_ON(txt_heap_base == 0);
--- a/xen/arch/x86/intel_txt.c
+++ b/xen/arch/x86/intel_txt.c
@@ -5,37 +5,12 @@
 #include <xen/string.h>
 #include <asm/page.h>
 #include <asm/intel_txt.h>
+#include <asm/slaunch.h>
 #include <xen/init.h>
 #include <xen/mm.h>
 #include <xen/slr_table.h>
 
 static uint64_t __initdata txt_heap_base, txt_heap_size;
-
-bool __initdata slaunch_active;
-
-static void __maybe_unused compile_time_checks(void)
-{
-    BUILD_BUG_ON(sizeof(slaunch_active) != 1);
-}
-
-int __init map_l2(unsigned long paddr, unsigned long size)
-{
-    unsigned long aligned_paddr = paddr & ~((1ULL << L2_PAGETABLE_SHIFT) - 1);
-    unsigned long pages = ((paddr + size) - aligned_paddr);
-    pages = ROUNDUP(pages, 1ULL << L2_PAGETABLE_SHIFT) >> PAGE_SHIFT;
-
-    if ( (aligned_paddr + pages * PAGE_SIZE) <= PREBUILT_MAP_LIMIT )
-        return 0;
-
-    if ( aligned_paddr < PREBUILT_MAP_LIMIT ) {
-        pages -= (PREBUILT_MAP_LIMIT - aligned_paddr) >> PAGE_SHIFT;
-        aligned_paddr = PREBUILT_MAP_LIMIT;
-    }
-
-    return map_pages_to_xen((unsigned long)__va(aligned_paddr),
-                            maddr_to_mfn(aligned_paddr),
-                            pages, PAGE_HYPERVISOR);
-}
 
 void __init map_txt_mem_regions(void)
 {
--- a/xen/arch/x86/intel_txt.c
+++ b/xen/arch/x86/intel_txt.c
@@ -15,11 +15,7 @@ static uint64_t __initdata txt_heap_base, txt_heap_size;
 
 void __init map_txt_mem_regions(void)
 {
-    void *evt_log_addr;
-    uint32_t evt_log_size;
-
     map_l2(TXT_PRIV_CONFIG_REGS_BASE, NR_TXT_CONFIG_PAGES * PAGE_SIZE);
-    map_l2(TPM_TIS_BASE, TPM_TIS_SIZE);
 
     txt_heap_base = read_txt_reg(TXTCR_HEAP_BASE);
     BUG_ON(txt_heap_base == 0);
@@ -28,16 +24,10 @@ void __init map_txt_mem_regions(void)
     BUG_ON(txt_heap_size == 0);
 
     map_l2(txt_heap_base, txt_heap_size);
-
-    find_evt_log(__va(slaunch_slrt), &evt_log_addr, &evt_log_size);
-    map_l2((unsigned long)evt_log_addr, evt_log_size);
 }
 
 void __init protect_txt_mem_regions(void)
 {
-    void *evt_log_addr;
-    uint32_t evt_log_size;
-
     uint64_t sinit_base, sinit_size;
 
     /* TXT Heap */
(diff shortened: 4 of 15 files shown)


xen/arch/x86/boot/head.S
```diff
@@ -522,21 +523,48 @@ __start:
...
 
 .Lslaunch_proto:
-        /* Save information that TrenchBoot slaunch was used. */
-        movb    $1, sym_esi(slaunch_active)
+        /* Upon reaching here, CPU state mostly matches the one setup by the
+         * bootloader with ESP, ESI and EDX being clobbered above. */
+
+        /* Save information that TrenchBoot slaunch was used and the value of
+         * EBP treating it as a parameter for slaunch. */
+        movb    $1,sym_esi(slaunch_active)
```
krystian-hebel: Apparently, a space between the operands is preferred in assembly, although this file uses both styles, and space was in minority of lines. Seeing as previous SL changes used mostly no-spaces style, I think we should leave it as it is and change it just before upstreaming.

xen/arch/x86/setup.c
```diff
@@ -1171,13 +1171,21 @@ void __init noreturn __start_xen(unsigned long mbi_p)
...
          * measured below by tpm_process_drtm_policy(). */
         tpm_measure_slrt();
-        /* Reserve TXT heap and SINIT. */
-        protect_txt_mem_regions();
+
+        if ( boot_cpu_data.x86_vendor == X86_VENDOR_INTEL )
+        {
+            /* Reserve TXT heap and SINIT. */
+            protect_txt_mem_regions();
+        }
```
krystian-hebel: SLB should also be mapped (in case it isn't located in memory pre-mapped by early Xen) and protected (it contains SLRT).

xen/arch/x86/smpboot.c
```diff
@@ -483,8 +483,8 @@ static int wakeup_secondary_cpu(int phys_apicid, unsigned long start_eip)
     if ( tboot_in_measured_env() && !tboot_wake_ap(phys_apicid, start_eip) )
         return 0;
 
-    if ( slaunch_active )
-        return slaunch_wake_aps(start_eip);
+    if ( slaunch_active && boot_cpu_data.x86_vendor == X86_VENDOR_INTEL )
```
krystian-hebel: `ap_boot_method` was created specifically for handling nonstandard SMP wake-ups, I think it should be used here as well. Not sure where would be proper place to set it, it isn't linked with early code, so perhaps somewhere around mapping and protecting TXT memory, unless you have better ideas.

xen/arch/x86/tpm.c
```diff
@@ -30,11 +30,25 @@ asm (
...
 #error "__va defined in non-paged mode!"
 #endif
 #define __va(x)     _p(x)
 
+/*
+ * Generic CPUID function
+ * clear %ecx since some cpus (Cyrix MII) do not set or clear %ecx
+ * resulting in stale register contents being returned.
+ */
+#define cpuid(_op,_eax,_ebx,_ecx,_edx)          \
```
krystian-hebel: I would put this in `#ifndef cpuid`, in case header files are reorganized, and move it to `slaunch.h` - it doesn't really have anything to do with TPM.

Also, I wonder if it would be better to do `is_intel_cpu()` analogous to AMD version in `txt_early()` instead of assembly. It's all the same to me, but some people would prefer to review C than assembly, which may make upstreaming slightly faster.

xen/arch/x86/tpm.c
```diff
@@ -961,8 +961,12 @@ void tpm_hash_extend(unsigned loc, unsigned pcr, uint8_t *buf, unsigned size,
 }
 
 #ifdef __EARLY_TPM__
-void __stdcall tpm_extend_mbi(uint32_t *mbi)
+void __stdcall tpm_extend_mbi(uint32_t *mbi, uint32_t param)
 {
+    /* Early TPM code isn't linked with the rest but still needs to have this
+     * variable holding correct value. */
+    slaunch_param = param;
```
krystian-hebel: I'll have to check if this links properly, early code is built differently.

SergiiDmytruk: There is second definition of `slaunch_param` at the top inside `#ifdef __EARLY_TPM__` to take this into account.

krystian-hebel: Ah, missed that, now it makes sense.

xen/arch/x86/boot/head.S
SergiiDmytruk: Kept the space: https://github.com/TrenchBoot/xen/compare/d99f5ae5f46d04d7044dfdf34a04370fda1b3417..4de56cbbf531cf9efb2efd4baf58f66db71b5ae9#diff-e4b4491b6c4e847098c559822abc0da4b62e431b9339a971a79026a0c3bac48aL531-R530

xen/arch/x86/setup.c
SergiiDmytruk: Mapped and protected: https://github.com/TrenchBoot/xen/compare/d99f5ae5f46d04d7044dfdf34a04370fda1b3417..4de56cbbf531cf9efb2efd4baf58f66db71b5ae9#diff-961c78e908063b25617eab27634027bfd5ba6bd6d9253c53a5ac322b94a7c791

xen/arch/x86/smpboot.c
SergiiDmytruk: Made use of it in d9176ce93703eb8d1c337b3d67e5e1c3d2042ff6.  Gets set in `xen/arch/x86/cpu/common.c`.

xen/arch/x86/tpm.c
SergiiDmytruk: e0d4647bc42c812f8bbb12152e39414c9e48dd0e. Moved to C renaming `txt_early` to `slaunch_early`.

xen/arch/x86/boot/head.S
```diff
@@ -522,18 +522,25 @@ __start:
...
 
 .Lslaunch_proto:
-        /* Save information that TrenchBoot slaunch was used. */
+        /* Upon reaching here, CPU state mostly matches the one setup by the
+         * bootloader with ESP, ESI and EDX being clobbered above. */
+
+        /* Save information that TrenchBoot slaunch was used and the value of
+         * EBP treating it as a parameter for slaunch. */
         movb    $1, sym_esi(slaunch_active)
+        movl    %ebp,sym_esi(slaunch_param)
```
krystian-hebel: `%ebp` is base address of SLB, but further code treats it as if it were SLRT. 16b offset to SLRT should be read from `6(%ebp)` and added to that base, but this may be dangerous to do on Intel - we don't know if it points to valid memory. Perhaps tuning it in `slaunch_early_tests()` would be the best option, that way we could pass SLRT location on Intel in the same variable, but it would require returning two variables (this is possible, SKL does it).

Another option would be to change SKL code to add this offset to `%ebp` just before jumping to Xen, but this would need to be explicitly mentioned in the specification. I believe I already mentioned need for describing the hand-off state in the issue with specification TODOs anyway.

xen/arch/x86/include/asm/slaunch.h
```diff
@@ -0,0 +1,96 @@
...
+ * resulting in stale register contents being returned.
+ *
+ * Copied from processor.h because that header can't be included by early code.
+ */
+#define cpuid(_op,_eax,_ebx,_ecx,_edx)          \
+    asm volatile ( "cpuid"                      \
+          : "=a" (*(int *)(_eax)),              \
+            "=b" (*(int *)(_ebx)),              \
+            "=c" (*(int *)(_ecx)),              \
+            "=d" (*(int *)(_edx))               \
```
krystian-hebel, line 35: Nit, maybe: by referencing memory locations returned values are always written to the memory, and then read from it to compare against vendor strings. This is suboptimal, especially since on Intel we run this code with UC memory. I don't know if there is a reason why Xen does it this way, perhaps @andyhhp does. Based on git history, it used to be inline function long time ago, which had to use pointers, but now this is a macro so it can modify whatever identifiers were passed to it in place.

That said, having suboptimal implementation is better than having two different implementations, and I don't expect you to fix every cpuid in the code as part of this PR.

xen/arch/x86/slaunch.c
```diff
@@ -0,0 +1,276 @@
...
+    BUILD_BUG_ON(SKINIT_SLB_SIZE != SKINIT_SLB_ALIGN);
+    BUILD_BUG_ON(SKINIT_SLB_SIZE == 0);
+    BUILD_BUG_ON((SKINIT_SLB_SIZE & (SKINIT_SLB_SIZE - 1)) != 0);
+
+    /* Rounding any address within SLB down to alignment gives SLB base and
+     * slaunch_param points to SLRT inside SLB on AMD. */
+    return slaunch_param & (SKINIT_SLB_SIZE - 1);
+}
+
+void __init map_skinit_mem_regions(void)
```
krystian-hebel: SLB is not enough, Xen needs to have at least TPM register space (0xfed40000-0xfed44fff) mapped:

```
(XEN) Early fatal page fault at e008:ffff82d0403133a1 (cr2=ffff8300fed40014, ec=0000)
(XEN) ----[ Xen-4.17.3  x86_64  debug=n  Not tainted ]----
(XEN) CPU:    0
(XEN) RIP:    e008:[<ffff82d0403133a1>] tpm_hash_extend+0x321/0x7f0
(XEN) RFLAGS: 0000000000010086   CONTEXT: hypervisor
(XEN) rax: ffff8300fed40014   rbx: ffff830003284000   rcx: 00000000fed40014
(XEN) rdx: 0000000000000000   rsi: 0000000000000002   rdi: ffff830000000000
(XEN) rbp: ffff82d0403e7d80   rsp: ffff82d0403e7b78   r8:  ffffffffffffffff
(XEN) r9:  ffff83000011e5bc   r10: 0000000000000000   r11: 0000000000008000
(XEN) r12: 0000000000000502   r13: ffff83000011e5bc   r14: 00000000009aa4a0
(XEN) r15: 0000000000400000   cr0: 0000000080050033   cr4: 00000000000000a0
(XEN) cr3: 00000000dbc43000   cr2: ffff8300fed40014
(XEN) fsb: 0000000000000000   gsb: 0000000000000000   gss: 0000000000000000
(XEN) ds: 0000   es: 0000   fs: 0000   gs: 0000   ss: 0000   cs: e008
(XEN) Xen code around <ffff82d0403133a1> (tpm_hash_extend+0x321/0x7f0):
(XEN)  d4 fe 48 09 c8 48 01 f8 <8b> 00 a9 00 00 00 50 0f 84 34 fe ff ff e9 f9 fd
```

In fact, it is also required on TXT, but because TXT register space is located in the same 2MB page as TPM, it went unnoticed.

krystian-hebel (commented):
Few more issues caught by testing on HW.

xen/arch/x86/include/asm/slaunch.h
andyhhp: This is out of date.  https://xenbits.xen.org/gitweb/?p=xen.git;a=commitdiff;h=86d0edda34e732f1b000d2adddf4f7884d18e04a finally purged the last of the macros for MISRA reasons.

In order to get code accepted into upstream Xen, it's going to need to be rebased onto the `staging` branch, and rearranged such that it can be integrated cleanly.

I'm happy to help out with whatever prep is necessary, and to provide some recommendations for how to structure the rest of the series.

Do you have a single branch with all work on it, rather than a delta against a still-pending delta?

xen/arch/x86/boot/head.S
SergiiDmytruk: > it would require returning two variables (this is possible, SKL does it).

SKL uses magic of SYSV ABI, while stdcall is forced in this case.  We can still return struct though, will just need to read it from memory.

xen/arch/x86/slaunch.c
SergiiDmytruk: 165c86c12102b283244173b63d169b20322fe534 and https://github.com/TrenchBoot/xen/compare/4de56cbbf531cf9efb2efd4baf58f66db71b5ae9..6c44eb9bcc4f73b6b98e6a6cfd50c0994c85580c#diff-1d266d1dff9daf02e019100b9c620a099561c98811a05a0b7581363b3c00ac39R59

xen/arch/x86/boot/head.S
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/4de56cbbf531cf9efb2efd4baf58f66db71b5ae9..6c44eb9bcc4f73b6b98e6a6cfd50c0994c85580c#diff-e4b4491b6c4e847098c559822abc0da4b62e431b9339a971a79026a0c3bac48a
and
https://github.com/TrenchBoot/xen/compare/4de56cbbf531cf9efb2efd4baf58f66db71b5ae9..6c44eb9bcc4f73b6b98e6a6cfd50c0994c85580c#diff-584be1f5aaa2f0679e692f6344b2e03670316e2ebdb7f4ba1dc4fbfb513d196c

Used output parameter instead of return value because returning structures seems to be grey area when it comes to ABI, so it's better to not rely on what compiler does.  In assembly the only change is passing address before all other parameters instead of after them, so it's pretty much the same either way.

xen/arch/x86/include/asm/slaunch.h
SergiiDmytruk: @krystian-hebel There are two such uses and each of them is invoked once resulting in 16 memory accesses in total, is that a lot?  If so, you might want to look at `objdump -d xen/arch/x86/boot/slaunch_early.o`, for example:
```asm
 615:   89 45 e8                mov    %eax,-0x18(%ebp)
 618:   8b 45 f4                mov    -0xc(%ebp),%eax
 61b:   89 44 24 10             mov    %eax,0x10(%esp)
 61f:   8b 45 0c                mov    0xc(%ebp),%eax
 622:   89 44 24 0c             mov    %eax,0xc(%esp)
 626:   8b 45 08                mov    0x8(%ebp),%eax
 629:   89 44 24 08             mov    %eax,0x8(%esp)
 62d:   8b 45 e8                mov    -0x18(%ebp),%eax
 630:   89 44 24 04             mov    %eax,0x4(%esp)
 634:   8b 45 ec                mov    -0x14(%ebp),%eax
 637:   89 04 24                mov    %eax,(%esp)
```
Early C code seems to be compiled without optimizations.

krystian-hebel: There is also implicit read done by `cmp` instructions. This is infinity times more than what is really necessary :slightly_smiling_face: In any case, as Andrew mentioned, we (will) have to abide to MISRA rules, so let's leave it as it is.

> In order to get code accepted into upstream Xen, it's going to need to be rebased onto the `staging` branch, and rearranged such that it can be integrated cleanly.

> Do you have a single branch with all work on it, rather than a delta against a still-pending delta?

We do not. Right now this is based on whatever Qubes OS uses, because this is what we're obligated to as part of [this project](https://docs.dasharo.com/projects/trenchboot-aem-v2/), any potential upstreaming is extra (as in: no deadline so lower priority).

xen/arch/x86/slaunch.c
```diff
@@ -0,0 +1,279 @@
...
+{
+    map_l2(TPM_TIS_BASE, TPM_TIS_BASE + TPM_TIS_SIZE);
+
+    map_l2(get_slb_start(), get_slb_start() + SKINIT_SLB_SIZE);
```
krystian-hebel: `map_l2()` takes size as a parameter, not the end address:

```suggestion
    map_l2(TPM_TIS_BASE, TPM_TIS_SIZE);

    map_l2(get_slb_start(), SKINIT_SLB_SIZE);
```

xen/arch/x86/intel_txt.c
```diff
@@ -5,44 +5,21 @@
...
-                            pages, PAGE_HYPERVISOR);
-}
-
 void __init map_txt_mem_regions(void)
 {
     void *evt_log_addr;
     uint32_t evt_log_size;
 
     map_l2(TXT_PRIV_CONFIG_REGS_BASE, NR_TXT_CONFIG_PAGES * PAGE_SIZE);
+    map_l2(TPM_TIS_BASE, TPM_TIS_BASE + TPM_TIS_SIZE);
```
krystian-hebel:
```suggestion
    map_l2(TPM_TIS_BASE, TPM_TIS_SIZE);
```

xen/arch/x86/boot/slaunch_early.c
```diff
@@ -127,5 +165,6 @@ uint32_t __stdcall txt_early_tests(uint32_t load_base_addr,
 
     verify_pmr_ranges(os_mle, os_sinit, load_base_addr, tgt_base_addr, size);
 
-    return os_mle->boot_params_addr;
+    result->mbi_pa = os_mle->boot_params_addr;
+    result->slaunch_param = slaunch_param;
```
krystian-hebel: Would returning `os_mle->slrt` here simplify later code? That way getting SLRT pointer would be the same for Intel and AMD. If that isn't a trivial change I'm OK with leaving it for further improvement, it isn't part of SKINIT support.

xen/arch/x86/intel_txt.c
SergiiDmytruk: Right... https://github.com/TrenchBoot/xen/compare/6c44eb9bcc4f73b6b98e6a6cfd50c0994c85580c..0eac7a817733f1f1d3a7ab853e49ad47473d0396#diff-31035528a674298c3fdbc80ca9452ac5a422984a24386556a0729ee18ab93aebR22

xen/arch/x86/slaunch.c
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/6c44eb9bcc4f73b6b98e6a6cfd50c0994c85580c..0eac7a817733f1f1d3a7ab853e49ad47473d0396#diff-1d266d1dff9daf02e019100b9c620a099561c98811a05a0b7581363b3c00ac39R58

xen/arch/x86/boot/slaunch_early.c
SergiiDmytruk: Not that it's a drastic simplification, but it's 26 less lines and, more importantly, more consistent behaviour between Intel and AMD.  https://github.com/TrenchBoot/xen/compare/6c44eb9bcc4f73b6b98e6a6cfd50c0994c85580c..0eac7a817733f1f1d3a7ab853e49ad47473d0396

xen/arch/x86/slaunch.c
krystian-hebel: You missed the SLB size.

SergiiDmytruk: Apparently this is the most complicated part of this PR. https://github.com/TrenchBoot/xen/compare/0eac7a817733f1f1d3a7ab853e49ad47473d0396..0bac2f30b7bb5c9f4946a0ea27f5d9fc816e80a0

krystian-hebel (commented):
Still not it, this time #PF when trying to read from event log:

<details>
<summary>Panic log</summary>

```
 Xen 4.17.3
(XEN) Xen version 4.17.3 (user@[unknown]) (gcc (GCC) 12.3.1 20230508 (Red Hat 12.3.1-1)) debug=n Thu Mar 21 19:37:56 UTC 2024
(XEN) Latest ChangeSet: 
(XEN) build-id: 82bfc42bf1b910810822c7134374e59fb4e605c4
(XEN) Bootloader: GRUB 2.06
(XEN) Command line: placeholder com1=115200,8n1 console=com1 loglvl=all dom0_mem=min:1024M dom0_mem=max:4096M ucode=scan smt=off gnttab_max_frames=2048 gnttab_max_maptrack_frames=4096
(XEN) Xen image load base address: 0xdb800000
(XEN) Video information:
(XEN)  VGA is text mode 80x25, font 8x16
(XEN)  VBE/DDC methods: none; EDID transfer time: 1 seconds
(XEN)  EDID info not retrieved because no DDC retrieval method detected
(XEN) Disc information:
(XEN)  Found 0 MBR signatures
(XEN)  Found 2 EDD information structures
(XEN) CPU Vendor: AMD, Family 23 (0x17), Model 1 (0x1), Stepping 2 (raw 00800f12)
(XEN) Xen-e820 RAM map:
(XEN)  [0000000000000000, 000000000000d4d3] (usable)
(XEN)  [000000000000d4d4, 000000000001d4d3] (reserved)
(XEN)  [000000000001d4d4, 00000000000997ff] (usable)
(XEN)  [0000000000099800, 000000000009ffff] (reserved)
(XEN)  [00000000000e0000, 00000000000fffff] (reserved)
(XEN)  [0000000000100000, 0000000076daffff] (usable)
(XEN)  [0000000076db0000, 0000000076ffffff] (reserved)
(XEN)  [0000000077000000, 00000000d8d5dfff] (usable)
(XEN)  [00000000d8d5e000, 00000000da209fff] (reserved)
(XEN)  [00000000da20a000, 00000000da373fff] (usable)
(XEN)  [00000000da374000, 00000000da727fff] (ACPI NVS)
(XEN)  [00000000da728000, 00000000dac66fff] (reserved)
(XEN)  [00000000dac67000, 00000000dbffffff] (usable)
(XEN)  [00000000dc000000, 00000000dfffffff] (reserved)
(XEN)  [00000000eff00000, 00000000eff7ffff] (reserved)
(XEN)  [00000000efff0000, 00000000efff0fff] (reserved)
(XEN)  [00000000fea00000, 00000000feafffff] (reserved)
(XEN)  [00000000fec00000, 00000000fec00fff] (reserved)
(XEN)  [00000000fec10000, 00000000fec10fff] (reserved)
(XEN)  [00000000fed00000, 00000000fed00fff] (reserved)
(XEN)  [00000000fed40000, 00000000fed44fff] (reserved)
(XEN)  [00000000fed80000, 00000000fed8ffff] (reserved)
(XEN)  [00000000fedc0000, 00000000fedc0fff] (reserved)
(XEN)  [00000000fedc2000, 00000000fedc8fff] (reserved)
(XEN)  [00000000fee00000, 00000000feefffff] (reserved)
(XEN)  [00000000ff000000, 00000000ffffffff] (reserved)
(XEN)  [0000000100000000, 000000101f37ffff] (usable)
(XEN)  [000000101f380000, 000000101fffffff] (reserved)
(XEN) Early fatal page fault at e008:ffff82d040313740 (cr2=ffff830003284038, ec=0000)
(XEN) ----[ Xen-4.17.3  x86_64  debug=n  Not tainted ]----
(XEN) CPU:    0
(XEN) RIP:    e008:[<ffff82d040313740>] tpm_hash_extend+0x720/0x790
(XEN) RFLAGS: 0000000000010046   CONTEXT: hypervisor
(XEN) rax: 0000000000000000   rbx: ffff830003284000   rcx: 0000000000000000
(XEN) rdx: 0000000000000000   rsi: 0000000000008000   rdi: ffff82d0403e7d58
(XEN) rbp: ffff82d0403e7d80   rsp: ffff82d0403e7b78   r8:  ffffffffffffffff
(XEN) r9:  ffff830000000000   r10: 0000000000000000   r11: 0000000000000502
(XEN) r12: ffff83000011d52c   r13: ffff83000011d5bc   r14: 00000000009aa4a0
(XEN) r15: 0000000000400000   cr0: 0000000080050033   cr4: 00000000000000a0
(XEN) cr3: 00000000dbc43000   cr2: ffff830003284038
(XEN) fsb: 0000000000000000   gsb: 0000000000000000   gss: 0000000000000000
(XEN) ds: 0000   es: 0000   fs: 0000   gs: 0000   ss: 0000   cs: e008
(XEN) Xen code around <ffff82d040313740> (tpm_hash_extend+0x720/0x790):
(XEN)  0f 1f 84 00 00 00 00 00 <8b> 43 38 48 89 c1 48 8d 44 83 3c 80 38 14 4c 8d
(XEN) Xen stack trace from rsp=ffff82d0403e7b78:
(XEN)    0000000000000001 ffff82d0405ae3a0 0000001100000400 ffff830000400000
(XEN)    ffff82d04024ae90 ffff83000011d5bc ffff82d000000013 ffff82d0403e7cd0
(XEN)    00000002009aa4a0 ffff82d04024e5da 00000010405ae3e5 ffff82d040445780
(XEN)    0000000000000082 0000000d40372ead ffff82d040372ea7 0000000000000006
(XEN)    ffff82d04024ae90 ffff82d04024e02b ffff82d040376b9e 0000000000000010
(XEN)    ffff82d04024ac40 ffff82d04024e5da 0000001000000002 ffff82d040445780
(XEN)    0000000000000082 0000000d0a376b9e 0000000000000000 ffff830000c00000
(XEN)    0000000000000106 ffff82d040443000 0000000001ffff06 1000000000000000
(XEN)    ffff82d0402ff458 ffff830000c00000 0000000000000030 00000000dbc3b063
(XEN)    ffff8300dbc41000 0000000000000063 ffff82e001b78820 1000000000000000
(XEN)    000000ffffffffff ffff830000e00000 ffff8300dbc3b030 0000000000000000
(XEN)    0000000000000000 0000000000000000 0000000000000000 0000000000000000
(XEN)    0000000000000000 0000000000000000 0000000000000000 0000000000000000
(XEN)    0000000000000000 0000000000000000 0000000000000000 0000000000000000
(XEN)    0000000000000000 0000000000000000 0000000000000000 0000000000000000
(XEN)    ffff83000011d5a4 ffff83000011d52c ffff83000011d5bc 00000000009aa4a0
(XEN)    0000000000400000 0000000000000002 ffff82d04030ceeb 0000000000000013
(XEN)    ffff82d0404189c0 ffff82d04037748d ffff830000000000 ffffffffffffffff
(XEN)    ffff82d0403b7410 ffff830000097ec0 ffff82d0403d2eef ffff830000097fb0
(XEN)    ffffffffffffffff dbbe7e5c00000000 0000000000000000 0000000000000000
(XEN) Xen call trace:
(XEN)    [<ffff82d040313740>] R tpm_hash_extend+0x720/0x790
(XEN)    [<ffff82d04024ae90>] S ns16550.c#ns16550_start_tx+0/0x80
(XEN)    [<ffff82d04024e5da>] S serial_puts+0xba/0x160
(XEN)    [<ffff82d04024ae90>] S ns16550.c#ns16550_start_tx+0/0x80
(XEN)    [<ffff82d04024e02b>] S serial.c#__serial_putc+0xcb/0x1b0
(XEN)    [<ffff82d04024ac40>] S ns16550.c#ns16550_putc+0/0x40
(XEN)    [<ffff82d04024e5da>] S serial_puts+0xba/0x160
(XEN)    [<ffff82d0402ff458>] S mm.c#virt_to_xen_l3e+0xb8/0x160
(XEN)    [<ffff82d04030ceeb>] S tpm_process_drtm_policy+0x2ab/0x3d0
(XEN)    [<ffff82d0403b7410>] S trampoline_realmode_entry+0/0x26
(XEN)    [<ffff82d0403d2eef>] S __start_xen+0x17af/0x273e
(XEN)    [<ffff82d04027379b>] S __high_start+0xdb/0xe0
(XEN) 
(XEN) Pagetable walk from ffff830003284038:
(XEN)  L4[0x106] = 00000000dbc41063 ffffffffffffffff
(XEN) 
(XEN) ****************************************
(XEN) Panic on CPU 0:
(XEN) FATAL TRAP: vec 14, #PF[0000] IN INTERRUPT CONTEXT
(XEN) ****************************************
(XEN) 
(XEN) Reboot in five seconds...
```

</details>

Perhaps TIS and event log should be mapped separately from TXT/SKINIT specific ranges, these are common to both versions, especially since SLRT is now in common variable.

SergiiDmytruk:
> Perhaps TIS and event log should be mapped separately from TXT/SKINIT specific ranges, these are common to both versions, especially since SLRT is now in common variable.

Good idea.  Also moved dispatching to `slaunch.c` out of `setup.c`.
https://github.com/TrenchBoot/xen/compare/0bac2f30b7bb5c9f4946a0ea27f5d9fc816e80a0..eb394884c439fca26cc5814cff59adfa10b21074

xen/arch/x86/slaunch.c
```diff
@@ -0,0 +1,316 @@
...
+{
+    /* The runtime computation relies on size being a power of 2 and equal to
+     * alignment. Make sure these assumptions hold. */
+    BUILD_BUG_ON(SKINIT_SLB_SIZE != SKINIT_SLB_ALIGN);
+    BUILD_BUG_ON(SKINIT_SLB_SIZE == 0);
+    BUILD_BUG_ON((SKINIT_SLB_SIZE & (SKINIT_SLB_SIZE - 1)) != 0);
+
+    /* Rounding any address within SLB down to alignment gives SLB base and
+     * SLRT is inside SLB on AMD. */
+    return slaunch_slrt & (SKINIT_SLB_SIZE - 1);
```
krystian-hebel: Another sneaky error that I missed:
```suggestion
    return slaunch_slrt & ~(SKINIT_SLB_SIZE - 1);
```

```
(XEN) SLAUNCH: reserving SLB (0xd4d4 - 0x1d4d4)
```

krystian-hebel (commented):
Now booting works all the way to dom0 starting, where it complains about E820 conflict ([log.txt](https://github.com/TrenchBoot/xen/files/14722127/log.txt)), which may be caused by SLB reservation.

There are also problems with AP bring-up, I think I know what's causing it (INIT is broadcasted, but it shouldn't be after SKINIT), but since I'll be reworking that code as part of SMP changes upstreaming, there is no reason to duplicate work here. dom0 is able to boot with 1 vCPU, albeit very slowly.

xen/arch/x86/slaunch.c
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/eb394884c439fca26cc5814cff59adfa10b21074..8ae81e157bfd4fa36c6bdb1dbf104d47edd5ce29

SergiiDmytruk:
> Now booting works all the way to dom0 starting, where it complains about E820 conflict ([log.txt](https://github.com/TrenchBoot/xen/files/14722127/log.txt)), which may be caused by SLB reservation.

No, it's TPM event log:
```
...
(XEN) SLAUNCH: reserving TPM event log (0x3284000 - 0x328c000)
...
(XEN)  Dom0 kernel: 64-bit, PAE, lsb, paddr 0x1000000 -> 0x5000000
...
```
Xen loads Dom0 kernel over a reserved area.  I've changed how GRUB allocates the log in https://github.com/TrenchBoot/grub/commit/89e27febfd0fe8bb0a52fd81f79d930949b67b83 and was able to boot.

That line should actually stay `SLAUNCH: reserving event log` (not "TPM event log") to not break `anti-evil-maid-dump-evt-log`.  Updated it in https://github.com/TrenchBoot/xen/compare/8ae81e157bfd4fa36c6bdb1dbf104d47edd5ce29..0186fd858e26497152cd9a27cb2a26c69217c156..

SergiiDmytruk:
Sent https://github.com/TrenchBoot/grub/pull/19 for that GRUB change.

SergiiDmytruk:
Looks like `0x1000000` is where Linux wants to get loaded:
<details><summary>readelf -l</summary>
<p>

```
Elf file type is EXEC (Executable file)
Entry point 0x1000000
There are 5 program headers, starting at offset 64

Program Headers:
  Type           Offset             VirtAddr           PhysAddr
                 FileSiz            MemSiz              Flags  Align
  LOAD           0x0000000000200000 0xffffffff81000000 0x0000000001000000
                 0x0000000001b5d72c 0x0000000001b5d72c  R E    0x200000
  LOAD           0x0000000001e00000 0xffffffff82c00000 0x0000000002c00000
                 0x00000000008c4000 0x00000000008c4000  RW     0x200000
  LOAD           0x0000000002800000 0x0000000000000000 0x00000000034c4000
                 0x0000000000035000 0x0000000000035000  RW     0x200000
  LOAD           0x00000000028f9000 0xffffffff834f9000 0x00000000034f9000
                 0x0000000001b07000 0x0000000001b07000  RWE    0x200000
  NOTE           0x0000000001d5d4e0 0xffffffff82b5d4e0 0x0000000002b5d4e0
                 0x000000000000024c 0x000000000000024c         0x8

 Section to Segment mapping:
  Segment Sections...
   00     .text .rodata .pci_fixup .tracedata .printk_index __ksymtab __ksymtab_gpl __ksymtab_strings __init_rodata __param __modver __ex_table .notes
   01     .data __bug_table .orc_unwind_ip .orc_unwind .orc_lookup .vvar
   02     .data..percpu
   03     .init.text .altinstr_aux .init.data .x86_cpu_dev.init .parainstructions .retpoline_sites .return_sites .altinstructions .altinstr_replacement .apicdrivers .exit.text .smp_locks .bss .brk .init.scratch
   04     .notes
```

</p>
</details>


So Xen can't do much here.  It's weird that Linux checks for e820 overlaps instead of Xen, but whatever, event log just should be allocated high to avoid this problem.

krystian-hebel:
Works (somewhat) with https://github.com/TrenchBoot/secure-kernel-loader/pull/17 and https://github.com/TrenchBoot/grub/commit/263639999fc9309ec721fd04cab0a053c98a6939. I've worked around the issue fixed with https://github.com/TrenchBoot/grub/pull/19 by calling `slaunch_module /skl.bin` before Xen and modules are loaded, which changed memory map sufficiently, I don't see how that fix can break anything so I'll merge it soon. I had to start with `maxcpus=1`, after which none of the `sys-*` VMs started due to timeouts. They wanted 2 vCPUs each, which was very inefficient with only one pCPU available. It was possible to start VMs manually after logging in through serial (remove `plymouth.ignore-serial-consoles` or something like that from Linux command line), one by one, after multiple tries.

Dumping event log works, but PCR 17 values don't match with event log. I'm not sure what's causing it, but I'm merging this PR as is. SMP issue will have to be resolved first, then we can worry about debugging PCR mismatches.

<details><summary>Logs</summary>

```
[user@dom0 ~]$ sudo anti-evil-maid-dump-evt-log                                 
Found event log at 0x4b8d000, size 0x8000 (32768) bytes                         
32768+0 records in                                                              
32768+0 records out                                                             
32768 bytes (33 kB, 32 KiB) copied, 1.29507 s, 25.3 kB/s                        
Found 2 algorithms:                                                             
    ID 0x0004 size = 0x14                                                       
    ID 0x000b size = 0x20                                                       
vendorInfoSize = 20                                                             
                                                                                
Entry 1:                                                                        
    PCR:        17                                                              
    Event Type: 0x502                                                           
    Digests:                                                                    
      SHA1: d3a60c696fd3be9d28916696c60a543265b69b87                            
      SHA256: 28d7f9982f249def7215719ebed463166ab14e17f6e1e26865a70232e48b01e8  
    Event: SKINIT                                                               
                                                                                
Entry 2:                                                                        
    PCR:        17                                                              
    Event Type: 0x502                                                           
    Digests:                                                                    
      SHA1: b34a6ceee3b376401e8cd1e21c735ccf60f3d216                            
      SHA256: 14fe6be01fc9b0f31bd1bf98929f2e41f010ecb7d9ba42a9c82458745facd649  
    Event: DLME entry offset                                                    
                                                                                
Entry 3:                                                                        
    PCR:        17                                                              
    Event Type: 0x502                                                           
    Digests:                                                                    
      SHA1: 21d3fd2a6e73a42e08ff9c554cd6ad2c9aef0b3a                            
      SHA256: ea2a4a1771e917547edf941daa49db0d29a6f986bed87c2fb9a71ddf8438d917  
    Event: DLME                                                                 
                                                                                
Entry 4:                                                                        
    PCR:        18                                                              
    Event Type: 0x502                                                           
    Digests:                                                                    
      SHA1: 4a00aca2a2ebbe387c4d685d04378e7b5928afa9                            
      SHA256: b61608d7a64209f51f2ed04f1d7cf398d7d9a91ddf5ce63cb90d39dd0df0755d  
    Event:                                                                      
                                                                                
Entry 5:                                                                        
    PCR:        17                                                              
    Event Type: 0x502                                                           
    Digests:                                                                    
      SHA1: c211668deab8c9b129a9e53b3b9c89274cee298f                            
      SHA256: 6d4e7be482a2383f2d94e11380a3b8b8b07222fc697c2aa830b2147b410d4ec0  
    Event: Measured MB2 module                                                  
                                                                                
Entry 6:                                                                        
    PCR:        17                                                              
    Event Type: 0x502                                                           
    Digests:                                                                    
      SHA1: 8a89b4c3384c25287b2eb165b20e4f0ffb248295                            
      SHA256: 52344d67244ac56d3409238f8759f5bd6944cebba1328838adebf19063a8fab8  
    Event: Measured MB2 module                                                  
                                                                                
[user@dom0 ~]$ sudo tpm2_pcrread                                                
  sha1:                                                                         
    0 : 0x3DCAEA25DC86554D94B94AA5BC8F735A49212AF8                              
    1 : 0xB2A83B0EBF2F8374299A5B2BDFC31EA955AD7236                              
    2 : 0x10BEA5C1802C8FF9AE280A100715B403A4F261FB                              
    3 : 0xB2A83B0EBF2F8374299A5B2BDFC31EA955AD7236                              
    4 : 0xB2A83B0EBF2F8374299A5B2BDFC31EA955AD7236                              
    5 : 0xB2A83B0EBF2F8374299A5B2BDFC31EA955AD7236                              
    6 : 0xB2A83B0EBF2F8374299A5B2BDFC31EA955AD7236                              
    7 : 0x08309BBB39E011058F8B55F6AC6F5A8317B0D726                              
    8 : 0x0000000000000000000000000000000000000000                              
    9 : 0x0000000000000000000000000000000000000000                              
    10: 0x9C1FCF0D800A677D0A27AF27FF4B157468DC4813                              
    11: 0x0000000000000000000000000000000000000000                              
    12: 0x0000000000000000000000000000000000000000                              
    13: 0x0000000000000000000000000000000000000000                              
    14: 0x0000000000000000000000000000000000000000                              
    15: 0x0000000000000000000000000000000000000000                              
    16: 0x0000000000000000000000000000000000000000                              
    17: 0x027DE1FDA0835D5AF195358892B7BDD4C709AA7D                              
    18: 0xFC534FF2A0B0863932B8AA715E812C42E4120DC6                              
    19: 0x0000000000000000000000000000000000000000                              
    20: 0x0000000000000000000000000000000000000000                              
    21: 0x0000000000000000000000000000000000000000                              
    22: 0x0000000000000000000000000000000000000000                              
    23: 0x0000000000000000000000000000000000000000                              
  sha256:                                                                       
    0 : 0x0000000000000000000000000000000000000000000000000000000000000000      
    1 : 0x0000000000000000000000000000000000000000000000000000000000000000      
    2 : 0x0000000000000000000000000000000000000000000000000000000000000000      
    3 : 0x0000000000000000000000000000000000000000000000000000000000000000      
    4 : 0x0000000000000000000000000000000000000000000000000000000000000000      
    5 : 0x0000000000000000000000000000000000000000000000000000000000000000      
    6 : 0x0000000000000000000000000000000000000000000000000000000000000000      
    7 : 0x0000000000000000000000000000000000000000000000000000000000000000      
    8 : 0x0000000000000000000000000000000000000000000000000000000000000000      
    9 : 0x0000000000000000000000000000000000000000000000000000000000000000      
    10: 0xBF0D858E3904704B36740BC2DDCF4820B93A9323C1098338B7C38E338735257B      
    11: 0x0000000000000000000000000000000000000000000000000000000000000000      
    12: 0x0000000000000000000000000000000000000000000000000000000000000000      
    13: 0xE70130ABD99D9920DA2698049EAB08F636A345B4D6E881E32CB1D1B05854BDF0      
    14: 0x0000000000000000000000000000000000000000000000000000000000000000      
    15: 0x0000000000000000000000000000000000000000000000000000000000000000      
    16: 0x0000000000000000000000000000000000000000000000000000000000000000      
    17: 0x402108CE794726A330FA50A4D956452E2A71A1F31781398363FDB26DFC031D4A      
    18: 0x92D91C6550557FAB751873FB3C6AB3D0CE83E89BAC58AA588ED09EB7A2DBE0C6      
    19: 0x0000000000000000000000000000000000000000000000000000000000000000      
    20: 0x0000000000000000000000000000000000000000000000000000000000000000      
    21: 0x0000000000000000000000000000000000000000000000000000000000000000      
    22: 0x0000000000000000000000000000000000000000000000000000000000000000      
    23: 0x0000000000000000000000000000000000000000000000000000000000000000
    ```
</details>

krystian-hebel (approved):

PR #12, opened 2024-04-04
x86/boot: read APIC base from register

krystian-hebel:
Some CPUs don't use default APIC base. Address in MSR is always valid, and it is already read to test for x2APIC.


krystian-hebel:
I've tested it on HP t630, but not on KGPE.

SergiiDmytruk (approved):
LGTM and KGPE-16 works on this branch too.

PR #19, opened 2024-10-14
Aem staging 14 10 24 rebase

krystian-hebel:
Changes beyond rebase (from `aem-4.17.4`, few changes aren't merged there yet) itself:
- changed the early code to use fastcall instead of stdcall
- updated comment on entry point, it was partially describing state on entry to ACM, not MLE (for some reason ACM entry is better described in Intel documentation than MLE)
- based on above, modified few first instructions accordingly (`%esi` calculated based on `%ebx` instead of call/pop, GDT loaded with CS selector as DS is undefined)
- changed `#include ""` to now supported `#include <>` in the early code, cleaned it up a bit
- made sure that every commit builds (for `git bisect`), but haven't test booting yet
- ~~somehow I messed up `tpm.c` during rebase - its early entry point lost an argument, but as it's called from assembly, no error was produced~~ never mind, my IDE didn't refresh the view automatically
- FIXME: `%ebp` is overwritten before it is passed to `slaunch_early_tests()`


krystian-hebel:
Modified the entry point to fix `%ebp` being overwritten, and made it more compatible with TXT. SKL will need to be changed accordingly.

There is `uint32_t slaunch_slrt;` in early TPM code, which will have to go in order to work with https://lore.kernel.org/xen-devel/20241007141539.1899350-1-frediano.ziglio@cloud.com/T/#m794a192186ed86a5a8adb51aaa45128a45885e5c. However, the same change will make it possible to link against external symbols, so the code will be able to access the final `slaunch_slrt` in `slaunch.c`, and local variable won't be needed at all. In fact, it could be set by `slaunch_early_tests()` and it won't have to be returned, which will further simplify arguments handling in `head.S`. Alternatively, `slaunch_early_tests()` may call `tpm_extend_mbi()` directly.

krystian-hebel:
Rebased to newer `staging`, renamed the target branch, but can't rename this one as it would close the PR.

Also included fix from https://github.com/TrenchBoot/xen/commit/d2aa2f5a5b4f506de30295737bbf285bc22eee94 (for others commits from https://github.com/TrenchBoot/xen/pull/17 I'd like to check on hardware whether they actually fix any issue before including them) and reordered the commits to group them into categories. The code still builds on each commit. Some commit messages became outdated, but additional ones will be outdated after 32b code linking, so I left them for now.

krystian-hebel:
Updated SLRT definition to the latest one. No rebase done today as the patch I'm waiting for hasn't been merged.

There is an ongoing unification of all possible Xen entry points (MB2, EFI, PVH) being done, which doesn't impact us too much, but slight modifications will be needed. https://xenbits.xen.org/gitweb/?p=xen.git;a=commit;h=0fe607b2a1440191b59cc6da83a3e717bf3ff7c0 would cause problems on old (pre-SLRT) implementation, but now everything is measured before `move_xen()`.

xen/arch/x86/boot/slaunch_early.c
```diff
@@ -104,13 +121,43 @@ static void verify_pmr_ranges(struct txt_os_mle_data *os_mle,
...
+        slrt = (struct slr_table *)result->slrt_pa;
+
+        result->mbi_pa = 0;
+        dl_info = (struct slr_entry_dl_info *)
+                  slr_next_entry_by_tag (slrt, NULL, SLR_ENTRY_DL_INFO);
+        /* Basic checks only, SKL checked and consumed the rest. */
+        if ( dl_info                             == NULL
+             || dl_info->hdr.size                != sizeof(*dl_info)
+             || dl_info->bl_context.bootloader   != SLR_BOOTLOADER_GRUB )
+            result->mbi_pa = dl_info->bl_context.context;
```
SergiiDmytruk: The condition is inverted.

xen/arch/x86/boot/head.S
```diff
@@ -486,20 +486,34 @@ __start:
...
+        sub     $8, %esp
+
+        push    %esp                             /* pointer to output structure */
         lea     sym_offs(__2M_rwdata_end), %ecx  /* end of target image */
         lea     sym_offs(_start), %edx           /* target base address */
         mov     %esi, %eax                       /* load base address */
-        /* txt_early_tests(load/eax, tgt/edx, tgt_end/ecx) using fastcall. */
-        call    txt_early_tests
+        /* slaunch_early_tests(load/eax, tgt/edx, tgt_end/ecx, ret/stk) using fastcall. */
+        call    slaunch_early_tests
```
SergiiDmytruk, line 535: Without `__stdcall` attribute on `slaunch_early_tests` this line must be followed by `add $4, %esp` to clean up the stack, which should be changed to `add $8, %esp` in the next commit (dad8074fc33891ad05f7073ac751de704c87791f).

xen/arch/x86/boot/head.S
```diff
@@ -368,12 +416,13 @@ __start:
...
 
+.Lset_stack:
         /* Set up stack. */
         lea     STACK_SIZE - CPUINFO_sizeof + sym_esi(cpu0_stack), %esp
 
         /* Initialize GDTR and basic data segments. */
-        add     %esi, sym_esi(gdt_boot_base)
-        lgdt    sym_esi(gdt_boot_descr)
+        add     %esi, %cs:sym_esi(gdt_boot_base)
+        lgdt    %cs:sym_esi(gdt_boot_descr)
```
SergiiDmytruk: Code segments are read-only, `add` causes segfault, so the second line doesn't need to change either.

SergiiDmytruk (dismissed):

xen/arch/x86/boot/head.S
krystian-hebel: I haven't thought about it. Then again, according to TXT Software Design Guide, on entry to MLE (Xen) DS, ES and SS are undefined. This makes it impossible to load new GDT without having the descriptor that would already have a proper base address in it...

Data segments were valid for all CPUs we've tested so far, so maybe the best option is to assume that this is an error in the documentation and ACM actually sets these segments. It has to somehow access the data anyway.

xen/arch/x86/boot/head.S
krystian-hebel: Right, this isn't true `__fastcall`:
https://github.com/TrenchBoot/xen/commit/b8aefd43c83f676ebf8e3f0405591ad8b5d0e2b5#diff-e4b4491b6c4e847098c559822abc0da4b62e431b9339a971a79026a0c3bac48aR511
https://github.com/TrenchBoot/xen/commit/d793572382cf94d64a25a4725370ea7f7439cc35#diff-e4b4491b6c4e847098c559822abc0da4b62e431b9339a971a79026a0c3bac48aR536

xen/arch/x86/boot/head.S
krystian-hebel: Undid that change and explained assumption in the comment: https://github.com/TrenchBoot/xen/commit/6d994f475c28b7b84fc89cf091ccc3599283ff4b#diff-e4b4491b6c4e847098c559822abc0da4b62e431b9339a971a79026a0c3bac48aR361-R365

xen/arch/x86/boot/slaunch_early.c
krystian-hebel: Nice catch, it would be difficult to debug.
https://github.com/TrenchBoot/xen/commit/d793572382cf94d64a25a4725370ea7f7439cc35#diff-584be1f5aaa2f0679e692f6344b2e03670316e2ebdb7f4ba1dc4fbfb513d196cR156

SergiiDmytruk (commented):
Works on APU with a small change on top to use amd_info instead of dl_info for newer SKL and GRUB.

`void *` -> `uint8_t *` I think goes somewhat against Qubes OS review.  Changes to not use things like `u32` and `__u32` in sha1/sha256 make files differ more from the original Linux sources.  Otherwise this version seem to be in alignment with the previous ones, although I didn't go line by line.

Mind that there is also a FIXME commit regarding stack_base.

krystian-hebel:
I've removed the FIXME commit, added smaller change in its place instead (https://github.com/TrenchBoot/xen/pull/19/commits/a994bec4157cbe7def06c8e41c7aceeee82c9c50). I've also moved change to x86/shutdown after commits directly related to SMP.

krystian-hebel:
https://lore.kernel.org/xen-devel/87d2a1e6-c69e-4789-b4e1-54aaf60eb858@amd.com/T/#m6c15a4ab1773d2563621532c3e324240d12a4134 got merged, I rebased this PR and updated the target branch to include it. This was the last of bootinfo patches.

SergiiDmytruk (approved):
> https://lore.kernel.org/xen-devel/87d2a1e6-c69e-4789-b4e1-54aaf60eb858@amd.com/T/#m6c15a4ab1773d2563621532c3e324240d12a4134 got merged

That's a nice change.  That place where they stored size in `mod_end` was really unexpected when I looked at the code.

PR #21, opened 2025-01-01
Slaunch support for Xen.efi

SergiiDmytruk:
An upstream fix will be gone on a future rebase.  Code improvements and measurement corrections follow.  The actual support is added in the penultimate commit, see its commit message for some details.  CI has some issues with the latest latest qubes-builderv2 (`sudo` isn't passwordless in a container which the builder creates), I just pinned its version to an older commit which works.

### DIFF
--- a/xen/arch/x86/boot/head.S
+++ b/xen/arch/x86/boot/head.S
@@ -540,7 +540,6 @@ __start:
         /* Push arguments to stack and call slaunch_early_tests(). */
         push    %esp                        /* pointer to output structure */
         push    %ebp                        /* Slaunch parameter on AMD */
-        push    %ebx                        /* Multiboot parameter */
         push    $sym_offs(__2M_rwdata_end)  /* end of target image */
         push    $sym_offs(_start)           /* target base address */
         push    %esi                        /* load base address */
--- a/xen/arch/x86/boot/head.S
+++ b/xen/arch/x86/boot/head.S
@@ -475,6 +475,13 @@ __pvh_start:
          *   which is supported by a given SINIT ACM
          */
 slaunch_stub_entry:
+#ifdef XEN_BUILD_EFI
+        mov     %ebx, %esi
+        sub     $sym_offs(slaunch_stub_entry), %esi
+        cmpb    $0, sym_esi(slaunch_efi_boot)
+        jne     slaunch_efi_jumpback
+#endif
+
         movl    $SLAUNCH_BOOTLOADER_MAGIC,%eax
 
         /* Fall through to Multiboot entry point. */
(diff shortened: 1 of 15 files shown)


xen/arch/x86/slaunch.c
```diff
@@ -142,17 +143,42 @@ void tpm_measure_slrt(void)
...
+            entry = (struct slr_entry_intel_info *)
+                slr_next_entry_by_tag(slrt, NULL, SLR_ENTRY_INTEL_INFO);
+            if ( entry == NULL )
                 panic("SLRT is missing Intel-specific information!\n");
 
-            tpm_hash_extend(DRTM_LOC, DRTM_DATA_PCR, (uint8_t *)intel_info,
-                            sizeof(*intel_info), DLE_EVTYPE_SLAUNCH, NULL, 0);
+            tmp = *entry;
+            tmp.boot_params_base = 0;
+            tmp.txt_heap = 0;
```
miczyg1, line 158: I don't quite understand why we have to zero these fields. The commit message does not explain anything.

xen/arch/x86/slaunch.c
```diff
@@ -142,17 +143,42 @@ void tpm_measure_slrt(void)
...
+
+            entry = (struct slr_entry_amd_info *)
+                slr_next_entry_by_tag(slrt, NULL, SLR_ENTRY_AMD_INFO);
+            if ( entry == NULL )
+                panic("SLRT is missing AMD-specific information!\n");
+
+            tmp = *entry;
+            tmp.next = 0;
+            tmp.slrt_base = 0;
+            tmp.boot_params_base = 0;
```
miczyg1, line 176: a/a

xen/arch/x86/slaunch.c
```diff
@@ -182,17 +208,43 @@ static struct slr_entry_policy *slr_get_policy(struct slr_table *slrt)
...
 
-static void check_drtm_policy(struct slr_table *slrt,
-                              struct slr_entry_policy *policy,
-                              struct slr_policy_entry *policy_entry,
-                              const struct boot_info *bi)
+static void check_slrt_policy_entry(struct slr_policy_entry *policy_entry,
+                                    struct slr_table *slrt)
+{
+    if ( policy_entry->entity_type != SLR_ET_SLRT )
+        panic("Second entry of DRTM policy in SLRT is not SLRT: %#04x!\n",
```
miczyg1: `SLRT is not SLRT` is a bit confusing. What does it mean?

miczyg1 (changes requested):

xen/arch/x86/slaunch.c
SergiiDmytruk: To prevent measurements from changing when the only thing that has changed is some address.  Addresses can vary due to bootloader, firmware or user doing something differently or just if GRUB gets bigger in size due to inclusion of more modules and ends up offsetting newly allocated memory.

The implementation in Linux does it and I repeat here even though boot params shouldn't even be set for Xen.

Updated commit message https://github.com/TrenchBoot/xen/pull/21/commits/33fe088f5dca429b628af16f460bcacf2c2c57f5, although it will probably be squashed later.

xen/arch/x86/slaunch.c
SergiiDmytruk: The first "SLRT" is related to DRTM policy which itself is stored within SLRT (`DRTM policy in SLRT`).  The policy has an entry describing measurement of SLRT.  I tried to avoid saying just "DRTM policy" as it provides too little context.  Simplified as this only harms the message.

https://github.com/TrenchBoot/xen/compare/f25f75bb48d3555bdc52df5afc2704763698046e..aa2fcc54193471e68cd2a730d5241e919128686b

miczyg1: Looks much better, thanks.

xen/arch/x86/boot/head.S
```diff
@@ -829,6 +829,107 @@ trampoline_setup:
...
+        mov     $BOOT_DS, %ecx
+        mov     %ecx, %ds
+        mov     %ecx, %es
+        mov     %ecx, %ss
+
+        push    $BOOT_CS32
+        lea     sym_esi(our_gdt),%edx
+        push    %edx
+        lret
+our_gdt:
```
krystian-hebel: This symbol can be hidden (`.Lour_gdt`). Alternatively local label (`1:`) can be used, `our_gdt` is a bit misleading.

xen/arch/x86/boot/head.S
```diff
@@ -829,6 +829,107 @@ trampoline_setup:
...
+
+        /*
+         * Pagetables are located in .bss and we are loaded to <4GB memory, so
```
krystian-hebel: Correct me if I'm wrong. The way I understand the code, `.bss` is cleared on first jump into Xen. Then it gets what it needs from UEFI, and jumps back to slaunch stub in GRUB, which in turn initiates DRTM. DCE gets measured and executed, then Xen code is measured and jumped into. At this point, `.bss` isn't cleared again.

What I'm getting at is that modified stub in GRUB can write to those higher dwords (thus allowing execution of code outside of measured part of Xen) without impacting PCR values.

I guess it applies to whole `.bss`, not just these pagetables, but they seem like the easiest target. I don't think we can clear `.bss` here, it may contain data obtained by functions called from `efi_start()`.

xen/arch/x86/slaunch.c
```diff
@@ -324,4 +374,52 @@ void tpm_process_drtm_policy(const struct boot_info *bi)
...
+        {
+            const struct boot_module *mod = &bi->mods[i];
+
+            paddr_t string = mod->cmdline_pa;
+            paddr_t start = mod->start;
+            size_t size = mod->size;
+
+            if ( mod->relocated || mod->released )
+            {
+                panic("Multiboot module \"%s\" (at %d) was consumed before measurement\n",
```
krystian-hebel: I'd drop `Multiboot` from here.

xen/arch/x86/slaunch.c
```diff
@@ -324,4 +374,52 @@ void tpm_process_drtm_policy(const struct boot_info *bi)
...
+             *
+             * 2 MiB is minimally mapped size and it should suffice.
+             */
+            map_l2(string, 2 * 1024 * 1024);
+            tpm_hash_extend(DRTM_LOC, DRTM_CODE_PCR,
+                            __va(string), strlen(__va(string)),
+                            DLE_EVTYPE_SLAUNCH, LOG_DATA("MB module string"));
+
+            map_l2(start, size);
+            tpm_hash_extend(DRTM_LOC, DRTM_DATA_PCR, __va(start), size,
```
krystian-hebel: `DRTM_CODE_PCR` and `DRTM_DATA_PCR` are reversed.

xen/arch/x86/boot/head.S
krystian-hebel: Perhaps moving https://github.com/TrenchBoot/xen/blob/aem-efi-boot/xen/common/efi/runtime.c#L67 to dedicated section would help.

krystian-hebel (changes requested):

xen/arch/x86/boot/head.S
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/aa2fcc54193471e68cd2a730d5241e919128686b..f484787ce7d2e2a1c0c4c7929ad9d8eeedcf7db3#diff-e4b4491b6c4e847098c559822abc0da4b62e431b9339a971a79026a0c3bac48aR851-R854

Renamed to `.Lgdt_is_set`.  Local labels are less readable.

xen/arch/x86/boot/head.S
SergiiDmytruk: Added clearing of bootmap:
 * https://github.com/TrenchBoot/xen/compare/aa2fcc54193471e68cd2a730d5241e919128686b..f484787ce7d2e2a1c0c4c7929ad9d8eeedcf7db3#diff-e4b4491b6c4e847098c559822abc0da4b62e431b9339a971a79026a0c3bac48aR856-R868
 * https://github.com/TrenchBoot/xen/compare/aa2fcc54193471e68cd2a730d5241e919128686b..f484787ce7d2e2a1c0c4c7929ad9d8eeedcf7db3#diff-4278d2e721441c82e835fb219fcc5a9da3ee669524cd33eb36a4939260a0909b

It's not even in `.bss`...  Most of initialization data is in `.init.data` or `.data.read_mostly`, however, something must  be in `.bss` because clearing it breaks the boot (no error messages and I didn't debug it).

By the way, I realized that `efi_arch_relocate_image()` uses bootmap after return from GRUB.  It seems to work, although this makes me wonder whether reusing them is safe:
```c
        /*
         * Relevant l{2,3}_bootmap entries get initialized explicitly in
         * efi_arch_memory_setup(), so we must not apply relocations there.
         * l2_directmap's first slot, otoh, should be handled normally, as
         * efi_arch_memory_setup() won't touch it (xen_phys_start should
         * never be zero).
         */
```

xen/arch/x86/slaunch.c
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/aa2fcc54193471e68cd2a730d5241e919128686b..f484787ce7d2e2a1c0c4c7929ad9d8eeedcf7db3#diff-1d266d1dff9daf02e019100b9c620a099561c98811a05a0b7581363b3c00ac39R403

xen/arch/x86/slaunch.c
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/aa2fcc54193471e68cd2a730d5241e919128686b..f484787ce7d2e2a1c0c4c7929ad9d8eeedcf7db3#diff-1d266d1dff9daf02e019100b9c620a099561c98811a05a0b7581363b3c00ac39R414-R419

SergiiDmytruk:
In addition to force-pushing changes (some of them are new for this PR) I changed the target branch to deal with 4.17.4 (which builder calls 4.17.5 for some reason) instead of staging.  This is because Xen staging has different ABI which can't be used on Qubes OS 4.2.

macpijan:
@miczyg1 all your threads are resolved here, can you please approve if you have no more comments?

miczyg1:
> @miczyg1 all your threads are resolved here, can you please approve if you have no more comments?

I can but there are also @krystian-hebel threads unresolved.

macpijan:
I am aware, @krystian-hebel will do his part as well.

miczyg1:
Also, I only tested the legacy boot path of this code and it didn't work so I am not convinced to give approve yet.

miczyg1 (approved):
@macpijan told we will be looking into legacy boot later too. For now I approve my review being addressed.

xen/arch/x86/efi/fixmlehdr.c
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    if ( fseek(fp, -(int)sizeof(header), SEEK_CUR) )
+    {
+        fprintf(stderr, "Failed to seek back again to MLE header in %s\n",
+                file_path);
+        goto fail;
+    }
+
+    header.entry_point += correction;
```
krystian-hebel, line 109: What about the size, shouldn't it be adjusted as well?

I think it would be better to modify MLE header definition in `head.S`, instead of patching it after build. There is `sym_offs()` macro that would do just that, but it probably would require some `#ifdef`s for non-UEFI case (unless the same offset has to be added there as well, in which case we were measuring wrong part of memory...).

xen/arch/x86/boot/head.S
```diff
@@ -475,6 +475,15 @@ __pvh_start:
          *   which is supported by a given SINIT ACM
          */
 slaunch_stub_entry:
+#ifdef XEN_BUILD_EFI
+        mov     %ebx, %esi
+        sub     $sym_offs(slaunch_stub_entry), %esi
+        cmpb    $0, sym_esi(slaunch_efi_boot)
+        je      1f
+        jmp     slaunch_efi_jumpback
+1:
```
krystian-hebel: I'm not sure if it is the case in Xen, but some reviewers are picky when it comes to local labels in conditionally compiled code.

```suggestion
        jne     slaunch_efi_jumpback
```

SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/f3db106684be8f16b54f75d1dccb23b1046c3d63..1278ae8b5f51de658b423849ccdfdb1a373acc8a

xen/arch/x86/efi/fixmlehdr.c
SergiiDmytruk: > What about the size, shouldn't it be adjusted as well?

Probably not.  ~~The size is computed as difference of offsets, so should be indifferent to consistent address shifts.~~ (EDIT: It's difference in both cases, but see https://github.com/TrenchBoot/grub/pull/28#discussion_r1962165890).

> I think it would be better to modify MLE header definition in `head.S`, instead of patching it after build. There is `sym_offs()` macro that would do just that, but it probably would require some `#ifdef`s for non-UEFI case (unless the same offset has to be added there as well, in which case we were measuring wrong part of memory...).

It would, but I ran out of ideas trying to make it work.  `head.S` is compiled only once (I think the same is true for most sources, some are compiled at most once), so no special casing.  Moving the header to separate files/sections in order to manipulate order or select correct one using linker scripts results in assembly errors.

krystian-hebel: Right, it seems that those differences are caused by linker only. These two cases are too different to make something common for both, as the base is counted differently between them (something like LMA and VMA). I guess we'll have to get back to it when working on testing on legacy again.

krystian-hebel (approved):

PR #23, opened 2025-03-23
Fix AEM for Intel TXT on UEFI platforms

SergiiDmytruk:
See commit messages for explanation of the changes.

This fixes https://github.com/TrenchBoot/trenchboot-issues/issues/56

### DIFF
--- a/xen/arch/x86/efi/efi-boot.h
+++ b/xen/arch/x86/efi/efi-boot.h
@@ -9,8 +9,15 @@
 #include <asm/microcode.h>
 #include <asm/msr.h>
 #include <asm/setup.h>
-#include <asm/intel_txt.h>
 #include <asm/slaunch.h>
+
+/*
+ * Make <asm/intel_txt.h> access TXT registers without address translation which
+ * is not set up at this point.
+ */
+#define __BOOT_DEFS_H__
+#include <asm/intel_txt.h>
+#undef __BOOT_DEFS_H__
 
 static struct file __initdata ucode;
 static uint64_t __initdata image_size;
--- a/xen/arch/x86/efi/efi-boot.h
+++ b/xen/arch/x86/efi/efi-boot.h
@@ -259,7 +259,7 @@ void __init noreturn start_xen_from_efi(void)
                 txt_os_sinit_data_start(txt_heap);
 
             txt_verify_pmr_ranges(os_mle, os_sinit, intel_info, xen_phys_start,
-                                  __XEN_VIRT_START, image_size);
+                                  xen_phys_start, image_size);
         }
     }
 
--- a/xen/arch/x86/include/asm/intel_txt.h
+++ b/xen/arch/x86/include/asm/intel_txt.h
@@ -393,8 +393,10 @@ static inline void txt_verify_pmr_ranges(struct txt_os_mle_data *os_mle,
          !is_in_pmr(os_sinit, tgt_base_addr, xen_size, check_high_pmr) )
         txt_reset(SLAUNCH_ERROR_LO_PMR_MLE);
 
-    /* Check if MBI is covered by PMR. MBI starts with 'uint32_t total_size'. */
-    if ( !is_in_pmr(os_sinit, info->boot_params_base,
+    /* Check if MBI is covered by PMR. MBI starts with 'uint32_t total_size'.
+     * There is no MBI on EFI boot. */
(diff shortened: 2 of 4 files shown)


xen/arch/x86/efi/efi-boot.h
```diff
@@ -3,6 +3,13 @@
  * is intended to be included by common/efi/boot.c _only_, and
  * therefore can define arch specific global variables.
  */
+
+/*
+ * Make <asm/intel_txt.h> access TXT registers without address translation which
+ * hasn't been set up yet.
+ */
+#define __BOOT_DEFS_H__
```
krystian-hebel: To avoid surprises in other headers, I'd put it immediately before `#include <asm/intel_txt.h>`, followed by `#undef __BOOT_DEFS_H__`.

xen/arch/x86/include/asm/intel_txt.h
```diff
@@ -388,11 +387,6 @@ static inline void txt_verify_pmr_ranges(struct txt_os_mle_data *os_mle,
     if ( !is_in_pmr(os_sinit, load_base_addr, xen_size, check_high_pmr) )
         txt_reset(SLAUNCH_ERROR_LO_PMR_MLE);
 
-    /* Check if all of Xen after relocation is covered by PMR. */
```
krystian-hebel: For legacy, relocation (if required) happens after this function is called, and Xen memory is measured before. This means that both ranges must be protected.

For UEFI, IIUC no relocation takes place after DL Event, so one check is sufficient there. The easiest way to achieve this would be to set `load_base_addr` and `tgt_base_addr` to the same value.

xen/arch/x86/slaunch.c
```diff
@@ -72,7 +99,7 @@ void __init map_slaunch_mem_regions(void)
         map_l2(get_slb_start(), SKINIT_SLB_SIZE);
     }
 
-    find_evt_log(__va(slaunch_slrt), &evt_log_addr, &evt_log_size);
+    find_evt_log(slr_get_table(), &evt_log_addr, &evt_log_size);
```
krystian-hebel, line 102: Truncated commit title?

krystian-hebel (changes requested):
What is the reason behind using `aem-next` based on v0.4.1 instead of `aem-staging-18-11-24`?

xen/arch/x86/efi/efi-boot.h
BeataZdunczyk: > What is the reason behind using aem-next based on v0.4.1 instead of aem-staging-18-11-24?

We should use the rebased version, considering all the effort we have put into it.

SergiiDmytruk: >> What is the reason behind using aem-next based on v0.4.1 instead of aem-staging-18-11-24?
>
> We should use the rebased version, considering all the effort we have put into it.

The latest Xen won't work because its ABI is incompatible with Qubes OS 4.2.  Changes were using rebased version at first, but those RPM packages wouldn't install.  I don't think we want to rebuild packages dependent on Xen and provide their AEM-specific versions to make latest Xen work or create a mixed package with 4.17.4 everything but latest Xen.EFI.

SergiiDmytruk: > To avoid surprises in other headers, I'd put it immediately before `#include <asm/intel_txt.h>`, followed by `#undef __BOOT_DEFS_H__`.

https://github.com/TrenchBoot/xen/compare/97a6affd50d6126c0b77e1e5ce88c0bb65823b14..1a33e907a021e6192b6fce3811fd977e44f3cf8e#diff-c66ef54b88eedfe71ad78fffd201179a093dbcf425532b0488817e8d457e9df9

xen/arch/x86/include/asm/intel_txt.h
SergiiDmytruk: https://github.com/TrenchBoot/xen/compare/97a6affd50d6126c0b77e1e5ce88c0bb65823b14..1a33e907a021e6192b6fce3811fd977e44f3cf8e

xen/arch/x86/slaunch.c
SergiiDmytruk: No, just ambiguous.  Reworded in 775ba698e7b926a4e72e1bf3e93772892b0333fa.

krystian-hebel (approved):

PR #26, opened 2026-04-10
.github: workflows: rebase-and-build.yml: add

DaniilKl:
Related to https://github.com/TrenchBoot/.github/pull/16 .

### DIFF
--- a/.github/workflows/rebase-build-and-publish-rebased.yml
+++ b/.github/workflows/rebase-build-and-publish-rebased.yml
@@ -1,0 +1,171 @@
+name: Rebase and build the last successful automatic rebase of aem-next-qubes branch
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: >
+          Set this input to do a dry run without building the packages to test
+          the rebase.
+        required: false
+        type: boolean
+        default: false
+  schedule:
+    - cron: '0 0 * * 6'
+
+concurrency:
+  group: automatic-rebase
+
+jobs:
+  prep-rebase:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout qubes-vmm-xen
+        uses: actions/checkout@v6
+        with:
+          repository: QubesOS/qubes-vmm-xen
+          path: qubes-vmm-xen
+      - name: Checkout downstream xen repository
+        uses: actions/checkout@v6
+        with:
+          repository: TrenchBoot/xen
+          token: ${{ secrets.TRENCHBOOT_REBASE_TOKEN }}
+          path: xen
+      - name: Read upstream version from qubes-vmm-xen
+        id: version
+        working-directory: qubes-vmm-xen
+        run: echo "version=$(cat version)" >> "$GITHUB_OUTPUT"
+      - name: Add upstream remote and fetch version tag
+        working-directory: xen
+        env:
+          UPSTREAM_TAG: RELEASE-${{ steps.version.outputs.version }}
+        run: |
+          git remote add upstream https://xenbits.xenproject.org/git-http/xen.git
+          git fetch upstream "refs/tags/${UPSTREAM_TAG}:refs/tags/${UPSTREAM_TAG}"
+      - name: Apply qubes-vmm-xen patches on top of upstream tag
+        working-directory: xen
+        env:
+          UPSTREAM_TAG: RELEASE-${{ steps.version.outputs.version }}
+        run: |
+          git checkout -b qubes-vmm-xen-with-patches-rebase-prep "$UPSTREAM_TAG"
+          SPEC="../qubes-vmm-xen/xen.spec.in"
+          mapfile -t PATCHES < <(grep -E '^Patch[0-9]+:' "$SPEC" | awk '{print $2}')
+          for patch_file in "${PATCHES[@]}"; do
+            git apply "../qubes-vmm-xen/${patch_file}"
+            escaped=$(printf '%s' "$patch_file" | sed 's/\./\\./g')
+            sed -i "/^Patch[0-9]*:[[:space:]]*${escaped}[[:space:]]*$/d" "$SPEC"
+            rm -f "../qubes-vmm-xen/${patch_file}"
+          done
+      - name: Copy QubesOS RPM files to downstream repository, rename spec.in
+        # Notes:
+        # 1. Renaming QubesOS/qubes-vmm-xen/config to
+        #    TrenchBoot/xen/config-qubesos, so it will not conflict with an
+        #    already existing directory "config" in the xen source code.
+        # 2. Renaming QubesOS/qubes-vmm-xen/xen.spec.in to
+        #    QubesOS/qubes-vmm-xen/vmm-xen.spec.in. This is done because the
+        #    qubes-dom0-packagev2.yml expects the file to have a name
+        #    ${{ inputs.qubes-component }}.spec.in and the correct name for this
+        #    component in QubesOS is vmm-xen, not xen.
+        run: |
+          mv qubes-vmm-xen/config qubes-vmm-xen/config-qubesos
+          sed -i 's/^Source3:[[:space:]]*config$/Source3: config-qubesos/' qubes-vmm-xen/xen.spec.in
+          cp -r qubes-vmm-xen/* xen/
+          mv xen/xen.spec.in xen/vmm-xen.spec.in
+          cd xen
+          git add -A
+          GIT_AUTHOR_NAME="github-actions[bot]" \
+          GIT_AUTHOR_EMAIL="github-actions[bot]@users.noreply.github.com" \
+          GIT_AUTHOR_DATE="2024-01-01T00:00:00" \
+          GIT_COMMITTER_NAME="github-actions[bot]" \
+          GIT_COMMITTER_EMAIL="github-actions[bot]@users.noreply.github.com" \
+          GIT_COMMITTER_DATE="2024-01-01T00:00:00" \
+          git commit --no-gpg-sign -m "Do not rebase: QubesOS patches, QubesOS RPM files and Qubes builder metadata"
+      - name: Push qubes-vmm-xen-with-patches branch to downstream
+        working-directory: xen
+        run: git push origin qubes-vmm-xen-with-patches-rebase-prep
+  count-skip-commits:
+    needs: prep-rebase
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      skip-commits: ${{ steps.count.outputs.skip-commits }}
+    steps:
+      - name: Checkout aem-next-qubes branch
+        uses: actions/checkout@v6
+        with:
+          repository: TrenchBoot/xen
+          ref: aem-next-qubes
+          fetch-depth: 0
+      - name: Fetch qubes-vmm-xen-with-patches-rebase-prep branch
+        run: git fetch origin qubes-vmm-xen-with-patches-rebase-prep
+      - name: Count commits with "Do not rebase:" after merge-base with qubes-vmm-xen-with-patches-rebase-prep
+        id: count
+        run: |
+          count=$(git rev-list --count --grep='Do not rebase:' origin/qubes-vmm-xen-with-patches-rebase-prep..HEAD)
+          echo "skip-commits=${count}" >> "$GITHUB_OUTPUT"
+  try-rebase:
+    needs: count-skip-commits
+    uses: TrenchBoot/.github/.github/workflows/rebase.yml@v2
+    secrets:
+      first-remote-token: ${{secrets.TRENCHBOOT_REBASE_TOKEN}}
+    permissions:
+      # For creation/deletion/pushing to branches and creating PRs
+      contents: write
+    with:
+      downstream-repo: 'https://github.com/TrenchBoot/xen.git'
+      downstream-branch: 'aem-next-qubes'
+      upstream-repo: 'https://github.com/TrenchBoot/xen.git'
+      upstream-branch: 'qubes-vmm-xen-with-patches-rebase-prep'
+      commit-user-name: 'github-actions[bot]'
+      commit-user-email: 'github-actions[bot]@users.noreply.github.com'
+      cicd-trigger-resume: '7. Rerun all jobs for the workflow https://github.com/TrenchBoot/xen/actions/runs/${{ github.run_id }} to resume automated rebase.'
+      skip-commits: ${{ needs.count-skip-commits.outputs.skip-commits }}
+      additional-params: '--use-commit-as-base'
+  cleanup-after-rebase-attempt:
+    needs: try-rebase
+    if: always()
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout downstream xen repository
+        uses: actions/checkout@v6
+        with:
+          repository: TrenchBoot/xen
+          token: ${{ secrets.TRENCHBOOT_REBASE_TOKEN }}
+          path: xen
+      - name: Delete qubes-vmm-xen-with-patches branch from downstream
+        working-directory: xen
+        env:
+          TOKEN: ${{ secrets.TRENCHBOOT_REBASE_TOKEN }}
+        run: |
+          git push "https://${TOKEN}@github.com/TrenchBoot/xen.git" \
+            --delete qubes-vmm-xen-with-patches-rebase-prep
+  get-version:
+    runs-on: ubuntu-latest
+    needs: try-rebase
+    if: ${{ needs.try-rebase.outputs.rebase-exit-code == '0' && ! inputs.dry_run }}
+    outputs:
+      version: ${{ steps.read-version.outputs.version }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: 'aem-next-qubes-rebased'
+      - name: Read version of the QubesOS Component from version file
+        id: read-version
+        # The aem-next-qubes-rebased should already have the version file either created
+        # in the prep-rebase or try-rebase jobs (it will be probably created by
+        # the prep-rebase as this file is a part of QubesOS repository) on which
+        # it depennds:
+        run: echo "version=$(cat version)" >> "$GITHUB_OUTPUT"
+  qubes-dom0-package:
+    needs: get-version
+    uses: TrenchBoot/.github/.github/workflows/qubes-dom0-packagev2.yml@v2
+    with:
+      qubes-component: 'vmm-xen'
+      qubes-component-branch: 'aem-next-qubes-rebased'
+      qubes-pkg-src-dir: '.'
+      qubes-pkg-version: ${{ needs.get-version.outputs.version }}


.github/workflows/rebase-and-build.yml
```diff
@@ -0,0 +1,41 @@
+name: Rebase on top of QubesOS main and build RPMs
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * 6'
+
+jobs:
+  try-rebase:
+    uses: TrenchBoot/.github/.github/workflows/rebase.yml@add-rebase-script
```
DaniilKl:
```suggestion
    uses: TrenchBoot/.github/.github/workflows/rebase.yml@master
```

This change must be applied after all the testing is done.

.github/workflows/rebase-and-build.yml
```diff
@@ -0,0 +1,41 @@
...
+  qubes-dom0-package:
+    uses: TrenchBoot/.github/.github/workflows/qubes-dom0-package.yml@master
+    with:
+      base-commit: 'e2cb10f236c83f1f1d78c04cc236e2935176f282'
+      patch-start: 1300
+      qubes-component: 'vmm-xen'
+      spec-pattern: '/^Patch1202:/'
+  trigger-gitea-cicd:
+    needs: antievilmaid
+    uses: TrenchBoot/.github/.github/workflows/trigger-gitea-workflow.yml@add-rebase-script
```
DaniilKl:
```suggestion
    uses: TrenchBoot/.github/.github/workflows/trigger-gitea-workflow.yml@master
```

This change must be applied after all the testing is done.

.github/workflows/rebase-and-build.yml
```diff
@@ -0,0 +1,41 @@
...
+      # For creation/deletion/pushing to branches and creating PRs
+      contents: write
+    with:
+      downstream-repo: 'https://github.com/trenchboot/xen.git'
+      downstream-branch: 'aem-next'
+      upstream-repo: 'https://github.com/QubesOS/qubes-vmm-xen.git'
+      upstream-branch: 'main'
+      commit-user-name: 'github-actions[bot]'
+      commit-user-email: 'github-actions[bot]@users.noreply.github.com'
+      cicd-trigger-resume: '7. Rerun the workflow https://github.com/TrenchBoot/qubes-antievilmaid/actions/runs/${{ github.run_id }} to resume automated rebase.'
```
DaniilKl: The rebasing for the `xen` component wont work this way, as the QubesOS changes are stored not as commits but as patches from the https://github.com/QubesOS/qubes-vmm-xen.git repository . Hence the flow should be similar to:

#### First job

1. Clone the xen upstream repository.
2. Apply the patches from the https://github.com/QubesOS/qubes-vmm-xen.git on the appropriate version of `xen` using `git am`.
3. Push it to the https://github.com/trenchboot/xen.git under appropriate branch.

#### Second job

1. Call the `rebase.yml` with the branch from the first job.

#### Third job

1. Call `qubes-dom0-packagev2.yml` so it will build RPM from a branch created in the second job.

#### Fourth job

1. Trigger Gitea CI/CD for signing.

DaniilKl: Though, the patches seems to not apply fully automatically. E.g., when applying the patch `0627-x86-mwait-idle-Use-ACPI-for-CPUs-without-hardcoded-C.patch` the `git am` fails with:

```
Applying: se.com/
fatal: empty ident name (for <>) not allowed
```

😕 .

.github/workflows/rebase-and-build.yml
DaniilKl: e87d2611c2ce596c8d5115641a256fbae5cc07e2

.github/workflows/rebase-and-build.yml
DaniilKl: e87d2611c2ce596c8d5115641a256fbae5cc07e2

.github/workflows/rebase-and-build.yml
DaniilKl: e87d2611c2ce596c8d5115641a256fbae5cc07e2

.github/workflows/rebase.yml
```diff
@@ -0,0 +1,106 @@
...
+        env:
+          UPSTREAM_TAG: RELEASE-${{ steps.version.outputs.version }}
+        run: |
+          git config user.name 'github-actions[bot]'
+          git config user.email 'github-actions[bot]@users.noreply.github.com'
+          git checkout -b qubes-vmm-xen-with-patches-rebase-prep "$UPSTREAM_TAG"
+          SPEC="../qubes-vmm-xen/xen.spec.in"
+          mapfile -t PATCHES < <(grep -E '^Patch[0-9]+:' "$SPEC" | awk '{print $2}')
+          for patch_file in "${PATCHES[@]}"; do
+            git am "../qubes-vmm-xen/${patch_file}"
```
DaniilKl:
```suggestion
            git am --committer-date-is-author-date --no-gpg-sign "../qubes-vmm-xen/${patch_file}"
```

Otherwise the recursion (Can it actually be called "recursion"? I just do not have another word for it.) here will not work.

DaniilKl: 7f4daed371a0fab7e0b737c60127fd297ae911e3

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,137 @@
...
+        id: version
+        run: echo "version=$(tr -d '[:space:]' < qubes-vmm-xen/version)" >> "$GITHUB_OUTPUT"
+      - name: Add upstream remote and fetch version tag
+        working-directory: xen
+        env:
+          UPSTREAM_TAG: RELEASE-${{ steps.version.outputs.version }}
+        run: |
+          git remote add upstream https://xenbits.xenproject.org/git-http/xen.git
+          git fetch upstream "refs/tags/${UPSTREAM_TAG}:refs/tags/${UPSTREAM_TAG}"
+      - name: Fix malformed patch header (split URL line in 0627 patch)
```
m-iwanicki: Maybe `sed -n '/^From: /,$p' | git am` for every patch to just remove everything before `From:` would work? As those parts are lost anyway during commit. That way you won't have to add hardcoded workarounds for single patch. Not only that but this patch name or number can change (e.g. it's 0268 on Trenchboot fork)

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,137 @@
...
+    steps:
+      - name: Checkout qubes-vmm-xen
+        uses: actions/checkout@v6
+        with:
+          repository: QubesOS/qubes-vmm-xen
+          path: qubes-vmm-xen
+      - name: Checkout downstream xen repository
+        uses: actions/checkout@v6
+        with:
+          repository: trenchboot/xen
```
m-iwanicki:
```suggestion
          repository: TrenchBoot/xen
```

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,137 @@
...
+          mapfile -t PATCHES < <(grep -E '^Patch[0-9]+:' "$SPEC" | awk '{print $2}')
+          for patch_file in "${PATCHES[@]}"; do
+            git am --committer-date-is-author-date --no-gpg-sign "../qubes-vmm-xen/${patch_file}"
+            escaped=$(printf '%s' "$patch_file" | sed 's/\./\\./g')
+            sed -i "/^Patch[0-9]*:[[:space:]]*${escaped}[[:space:]]*$/d" "$SPEC"
+            rm -f "../qubes-vmm-xen/${patch_file}"
+          done
+      - name: Copy QubesOS RPM files to downstream repository, rename spec.in
+        run: |
+          mv qubes-vmm-xen/config qubes-vmm-xen/config-qubesos
```
m-iwanicki, line 72: why rename? Maybe short comment?

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,137 @@
+name: Rebase and build the last successful automatic rebase of aem-next branch
+
+on:
+  workflow_dispatch:
```
m-iwanicki, line 4: Maybe add dry_run flag that'll do rebase but skip bulding/publishing? So we can at least partially test this in the future?

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,137 @@
...
+  schedule:
+    - cron: '0 0 * * 6'
```
m-iwanicki, line 14: I wonder if cron + workflow_dispatch won't conflict (race condition). Maybe add [concurrency](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-using-concurrency-and-the-default-behavior)?

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,137 @@
...
+    secrets:
+      woodpecker-token: ${{ secrets.WOODPECKER_TOKEN }}
+    with:
+      api-url:  'https://ci.3mdeb.com'
+      owner:    'zarhus'
+      repo:     'trenchboot-release-cicd-pipeline'
+      ref:      'master'
+      inputs: >-
+        --input GITHUB_REPO=xen
+        --input GITHUB_SHA=${{ github.sha }}
```
m-iwanicki: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#schedule

for schedule/cron this will be HEAD of the default branch (master)

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,137 @@
...
+            git am --committer-date-is-author-date --no-gpg-sign "../qubes-vmm-xen/${patch_file}"
+            escaped=$(printf '%s' "$patch_file" | sed 's/\./\\./g')
+            sed -i "/^Patch[0-9]*:[[:space:]]*${escaped}[[:space:]]*$/d" "$SPEC"
+            rm -f "../qubes-vmm-xen/${patch_file}"
+          done
+      - name: Copy QubesOS RPM files to downstream repository, rename spec.in
+        run: |
+          mv qubes-vmm-xen/config qubes-vmm-xen/config-qubesos
+          sed -i 's/^Source3:[[:space:]]*config$/Source3: config-qubesos/' qubes-vmm-xen/xen.spec.in
+          cp -r qubes-vmm-xen/* xen/
```
m-iwanicki, line 74: What if you replace some already existing, important files (maybe not currently but in the future)?

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,137 @@
...
+          mv xen/xen.spec.in xen/vmm-xen.spec.in
+          cd xen
+          git add -A
+          git commit -m "QubesOS RPM files and Qubes builder metadata"
+      - name: Push qubes-vmm-xen-with-patches branch to downstream
+        working-directory: xen
+        run: git push origin qubes-vmm-xen-with-patches-rebase-prep
```
m-iwanicki, line 87: I'm assuming this will fail if this branch already exists and `cleanup-after-rebase-attempt` didn't run (because previous job failed)

m-iwanicki (changes requested):
+ Pass commit message through a spellchecker

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,137 @@
...
+    needs: try-rebase
+    outputs:
+      version: ${{ steps.read-version.outputs.version }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: 'aem-next-rebased'
+      - name: Read version of the QubesOS Component from version file
+        id: read-version
+        run: echo "version=$(cat version)" >> $GITHUB_OUTPUT
```
m-iwanicki:
```suggestion
        run: echo "version=$(cat version)" >> "$GITHUB_OUTPUT"
```

actionlint suggestion

m-iwanicki (changes requested):

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,137 @@
+name: Rebase and build the last successful automatic rebase of aem-next branch
```
m-iwanicki: I'm not sure if we want Qubes patches on aem-next branch? We are/were using this branch for e.g. upstreaming Trenchboot changes. Maybe separate `qubes+aem-next` branch?

DaniilKl:
> * Pass commit message through a spellchecker

Done.

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: I know. This will not be an issue, because this value is being used to reconstruct the artifacts name on the Woodpecker side so it can download the artifacts from GitHub Actions page. The initial artifact name is built here: https://github.com/TrenchBoot/.github/blob/0de7bf67b7290eef7d533ba906b6c90caafe2690/.github/workflows/qubes-dom0-packagev2.yml#L208 in the following way:

```yaml
      - name: Save built packages
        uses: actions/upload-artifact@v4
        with:
          if-no-files-found: error
          name: qubesos.dom0.fc37-${{ inputs.qubes-component }}-${{ github.sha }}
          path: '*.rpm'
```

So in case the `github.sha` will equal to some unrelated to the `aem-next-rebased` branch -  the `qubes-dom0-packagev2.yml` will still add it to the artifact name on the GitHub Actions page and the CI/CD on Woodpecker side will still be able to download it.

On the other hand this will result the artifact name on the GitHub Actions page to be always the same as it will contain the commit hash of the default branch that seems to be rarely updated in some of the TrenchBoot repos. But this will not be a problem for the Woodpecker to download it, as during download you should specify the run ID which is unique even if it uses the `.yaml` workflow file from the same branch.

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: I am not sure too, lets wait for the maintainers to share their opinion: https://matrix.to/#/!WHWPvnIGPhGGtUFucJ:matrix.org/$N7FMWWIoHr1YYAv2iUcKaVgffxHm6C3ekj76h6xYQpo?via=matrix.org&via=matrix.tu-berlin.de&via=invisiblethingslab.com .

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: The `From:` and the things that come after it (the name and the email of the commit author) are not being lost during commit, it turns into `author` field instead. Consider the following example:

<details><summary>Details</summary>
<p>

```bash
danillklimuk in ~/test on master λ git ci --allow-empty -m "WIP1"
[master (root-commit) 322eb8aa34fd] WIP1
danillklimuk in ~/test on master ● λ git ci --allow-empty -m "WIP2"
[master 6394052b1639] WIP2
danillklimuk in ~/test on master λ git format-patch -1 HEAD
0001-WIP2.patch
danillklimuk in ~/test on master ● λ git reset --hard HEAD^
HEAD is now at 68f846c6f9d1 WIP2
danillklimuk in ~/test on master ● λ git config user.email 'some-bot'           
danillklimuk in ~/test on master ● λ git config user.name 'some-bot'
danillklimuk in ~/test on master ● λ git am ./0001-WIP2.patch 
danillklimuk in ~/test on master performing a am λ git am --allow-empty
Applying: WIP2
No changes - recorded it as an empty commit.
danillklimuk in ~/test on master ● λ git cat-file -p HEAD 
tree 4b825dc642cb6eb9a060e54bf8d69288fbee4904
parent 68f846c6f9d1e9e516efb74b5fe4d2057cb3751e
author github-actions[bot] <github-actions[bot]@users.noreply.github.com> 1776778356 +0200
committer some-bot <some-bot> 1776778408 +0200

WIP2

Signed-off-by: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
--
2.52.0

danillklimuk in ~/test on master ● λ cat 0001-WIP2.patch 
From 6394052b1639a136876390148ab3b9386e3e5a56 Mon Sep 17 00:00:00 2001
From: "github-actions[bot]" <github-actions[bot]@users.noreply.github.com>
Date: Tue, 21 Apr 2026 15:32:36 +0200
Subject: [PATCH] WIP2

Signed-off-by: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
-- 
2.52.0
```

</p>
</details>

m-iwanicki: > everything before `From:` [...] As those parts are lost

I meant, remove everything before `From:`.

DaniilKl: Makes sense: https://github.com/TrenchBoot/xen/compare/754e1eabfecfa1f16abf6d37bb1f7c0007ef3bc3..b29178a6dbec7a609604062f35fe8407d1c7ce14 .

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: https://github.com/TrenchBoot/xen/compare/754e1eabfecfa1f16abf6d37bb1f7c0007ef3bc3..b29178a6dbec7a609604062f35fe8407d1c7ce14

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: I had a little typo there: https://github.com/TrenchBoot/xen/compare/b29178a6dbec7a609604062f35fe8407d1c7ce14..7d083272adc7e78087879c980abfa744fd832dc2 .

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: I commented this one in the commit message, but on the second thought an inline comment will be useful too: https://github.com/TrenchBoot/xen/compare/754e1eabfecfa1f16abf6d37bb1f7c0007ef3bc3..b29178a6dbec7a609604062f35fe8407d1c7ce14 .

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: https://github.com/TrenchBoot/xen/compare/754e1eabfecfa1f16abf6d37bb1f7c0007ef3bc3..b29178a6dbec7a609604062f35fe8407d1c7ce14

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: Good point, I should have added the `always()`: https://github.com/TrenchBoot/xen/compare/754e1eabfecfa1f16abf6d37bb1f7c0007ef3bc3..b29178a6dbec7a609604062f35fe8407d1c7ce14 .

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: https://github.com/TrenchBoot/xen/compare/754e1eabfecfa1f16abf6d37bb1f7c0007ef3bc3..b29178a6dbec7a609604062f35fe8407d1c7ce14

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: https://github.com/TrenchBoot/xen/compare/7d083272adc7e78087879c980abfa744fd832dc2..9e91e1e21f1644662905e10d19fe096d283ac35c

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: The probability is rather low because of the names of the files: the first one contains `qubesos` and the second one is `.spec.in`.

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: Considering the answer: https://matrix.to/#/!WHWPvnIGPhGGtUFucJ:matrix.org/$Q_vxQEooVvzAs3ienpZNHQ8MMaxdyJ51cybIrmUjDvo?via=matrix.org&via=matrix.tu-berlin.de&via=invisiblethingslab.com, you are right, a separate branch for QubesOS things should be created and the auto-rebase should be run there, not on the aem-next branch.

The only thing that should be changed because of this fact in this PR - are the branches though, all the other things should stay untouched. So we can continue review, but this will be a bloker for the final merge.

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,155 @@
...
+    if: inputs.dry_run == ''
+    outputs:
+      version: ${{ steps.read-version.outputs.version }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: 'aem-next-rebased'
+      - name: Read version of the QubesOS Component from version file
+        id: read-version
+        run: echo "version=$(cat version)" >> "$GITHUB_OUTPUT"
```
SergiiDmytruk, line 163: No `tr` to remove whitespace here?  Why is it needed in other places?

DaniilKl: Actually I do not see a reason for the `tr` here as all the `version` files do not contain whitespaces: https://github.com/TrenchBoot/xen/pull/26/changes/4e4efd21cbe5fcecda8abe97c541c17270145ce1 .

.github/workflows/rebase-build-and-publish-rebased.yml
iwanicki92: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_dispatchinputsinput_idtype

I think boolean would be better?

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: Mea culpa, another typo: https://github.com/TrenchBoot/xen/compare/4e4efd21cbe5fcecda8abe97c541c17270145ce1..a12e0701c2eb58a0913e3463d4f756acd54c762f

.github/workflows/rebase-build-and-publish-rebased.yml
SergiiDmytruk: Maybe we can use CI variables to specify branches?  This way we don't have to hard-code names in commits and can change easily in the future.  Although, I'm not sure how often this will change, so it may not be worth it.

.github/workflows/rebase-build-and-publish-rebased.yml
m-iwanicki: I guess that's ok, in the worst case build will fail

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: https://github.com/TrenchBoot/xen/compare/bf52c84e81929a4b1f7f24b6f330142afad4d62c..ae91a9bac0ae3a17155f044417ee34a044103280

DaniilKl:
@m-iwanicki, @SergiiDmytruk, ping for review.

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: I will address this at the end, when the rebase logic will be accepted.

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,170 @@
...
+    permissions:
+      contents: read
+    outputs:
+      skip-commits: ${{ steps.count.outputs.skip-commits }}
+    steps:
+      - name: Checkout aem-next-test branch
+        uses: actions/checkout@v6
+        with:
+          repository: TrenchBoot/xen
+          ref: aem-next-test
```
m-iwanicki:
```suggestion
      - name: Checkout aem-next branch
        uses: actions/checkout@v6
        with:
          repository: TrenchBoot/xen
          ref: aem-next
```

I think this should be `aem-next`? (there is no `aem-next-test` branch)

m-iwanicki (changes requested):

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: https://github.com/TrenchBoot/xen/compare/cbd516687afe2523fa5d219b64299373f53f32ad..d7c372b308e4d4f11500f3ed57bd21d0dcfede74

Test leftovers.

.github/workflows/rebase-build-and-publish-rebased.yml
```diff
@@ -0,0 +1,170 @@
...
+        #    already existing directory "config" in the xen source code.
+        # 2. Renaming QubesOS/qubes-vmm-xen/xen.spec.in to
+        #    QubesOS/qubes-vmm-xen/vmm-xen.spec.in. This is done because the
+        #    qubes-dom0-packagev2.yml expects the file to have a name
+        #    ${{ inputs.qubes-component }}.spec.in and the correct name for this
+        #    component in QubesOS is vmm-xen, not xen.
+        run: |
+          mv qubes-vmm-xen/config qubes-vmm-xen/config-qubesos
+          sed -i 's/^Source3:[[:space:]]*config$/Source3: config-qubesos/' qubes-vmm-xen/xen.spec.in
+          cp -r qubes-vmm-xen/* xen/
```
m-iwanicki, line 74: Do you want to skip `.qubesbuilder` (and other dotfiles?)

```suggestion
          rsync -a --exclude='.git' qubes-vmm-xen/ xen/
```

m-iwanicki (changes requested):

.github/workflows/rebase-build-and-publish-rebased.yml
DaniilKl: https://github.com/TrenchBoot/xen/compare/d7c372b308e4d4f11500f3ed57bd21d0dcfede74..85acd908402dee22dd53988f7eae650c153affeb

The branch [aem-bext-qubes](https://github.com/TrenchBoot/xen/tree/aem-next-qubes) is a copy of the [aem-next](https://github.com/TrenchBoot/xen/tree/aem-next).

SergiiDmytruk: Hm, why make a copy?  I meant that the branch shouldn't be force-pushed automatically and having `aem-next-rebased` already achieves that.  Adding `aem-next-qubes` doesn't seem to add anything other than duplication unless `aem-next` can't be used as is.

While at it, we can also rename `aem-next` to `tb-dev` here for consistency with GRUB.

DaniilKl: > Hm, why make a copy? I meant that the branch shouldn't be force-pushed automatically and having `aem-next-rebased` already achieves that. Adding `aem-next-qubes` doesn't seem to add anything other than duplication unless `aem-next` can't be used as is.
> 
> While at it, we can also rename `aem-next` to `tb-dev` here for consistency with GRUB.

@SergiiDmytruk, ok, then I misunderstood you. Then I need to change the branches back. I assume you do now, that smbd will need to manage (e.g., force-push to the `aem-next`) the `aem-next-rebased` and delete it, otherwise the automation will not start the scheduled rebase and fail with  [code 6](https://github.com/TrenchBoot/.github/blob/56d15c416bbc266b8323021571363e26ade37c30/scripts/rebase.sh#L119).

SergiiDmytruk: Here's my Matrix comment:

> The two branches are somewhat closer to tracking QubesOS than upstream with changes for upstreaming having their own branches (`tb-dev` is on top of an upstream branch because unlike Xen this works in QubesOS).  That said, I'm also not sure that we want to autorebase them.  These are basically development branches created to facilitate pull requests (at different time, hence inconsistent naming).  We probably should have these as is for PRs, have autorebased versions for QubesOS and have upstream branches as separate entities which get created as needed.  Then `tb-dev`/`aem-next` can be synced with autorebased branches periodically.

I don't actually remember what I meant by it...  It can be interpreted differently depending on whether auto-rebasing targets only QubesOS packages or also development.  It's for QubesOS packages, right?  Then the current variant with 3 branches may actually be correct.  So the workflow is:
 1. Development happens on `aem-next`/`tb-dev`.
 2. At some point we push to `*-qubes` branch.
 3. If we like `*-qubes-rebased`, we replace `*-qubes` with it.
 4. At some later point we replace `aem-next`/`tb-dev` with `*-qubes`.

The only difference between 2 or 3 branches is whether there is step 4 or not.

Let's keep the configuration as is and update later if necessary.  It's not clear which is actually better until we have it working.  Sorry for the confusion.  Can also keep `aem-next` to not bother renaming in configuration and documentation, it's not that a big deal.

m-iwanicki (approved):

SergiiDmytruk (approved):

PR #27, opened 2026-04-15
xen/arch/x86: fix TXT AP gate placement and MONITOR/MWAIT race

accek-itl:
Two fixes for the TXT AP gate needed after upstream changes:
1. Gate TXT AP startup earlier: the gate must be before the cpu_info.cr4 write in early assembly.
2. Fix potential deadlock in MONITOR/MWAIT TXT AP gate loop.

Tested by booting Trenchboot-XEN on Asus LCL laptop ([aem-staging-2026-03-13-accek branch](https://github.com/accek-itl/xen/tree/aem-staging-2026-03-13-accek) which included these commits) and verifying PCRs 17, 18 are non-zero.

Related issue: https://github.com/TrenchBoot/trenchboot-issues/issues/82

### DIFF
--- a/xen/arch/x86/boot/x86_64.S
+++ b/xen/arch/x86/boot/x86_64.S
@@ -30,7 +30,10 @@ ENTRY(__high_start)
         cmp     %esp, (%rcx, %rax, 4)
         jne     1b
 
-        /* %eax is now Xen CPU index. */
+        mov     %esp, %edx
+
+        /* %eax is now Xen CPU index, %edx is APIC ID. */
+
         lea     stack_base(%rip), %rcx
         mov     (%rcx, %rax, 8), %rsp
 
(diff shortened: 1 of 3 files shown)


m-iwanicki:
Tested on DELL Optiplex 7010 (UEFI), Xen + Trenchboot works and boots, PCRs 17 and 18 are extended

accek:
I think for upstreaming these changes could be squashed into this commit: https://github.com/TrenchBoot/xen/commit/8b7678638a7ff28b7952ae2bb11145eb3637e3d2, though this commit message part will no longer hold:

```
With this patch, every AP goes through assembly part, and only when in
start_secondary() in C they re-enter MONITOR/MWAIT iff they are not the
AP that was asked to boot. The same address is reused for simplicity,
and on next wakeup call APs don't have to go through assembly part
again (GDT, paging, stack setting).
```

xen/arch/x86/boot/x86_64.S
```diff
@@ -40,6 +43,19 @@ ENTRY(__high_start)
...
         add     $(STACK_SIZE - CPUINFO_sizeof), %rsp
 
+        /*
+         * TXT AP gate.
+         *
+         * In TXT boot, SINIT releases all APs at once and they race into
+         * __high_start in parallel, so we gate non-target APs.
+         *
+         * In non-TXT boot APs wake one-by-one via SIPI.
+         */
```
SergiiDmytruk, line 59: I would explain that this serializes APs initialization to force them waking up in accordance with the order in which BSP expects them to wake up.  Could also mention `STACK_CPUINFO_FIELD()` below which necessitates doing it here.

xen/arch/x86/smpboot.c
```diff
@@ -240,6 +240,30 @@ static void smp_callin(void)
         cpu_relax();
 }
 
+unsigned int txt_booting_apicid;
```
SergiiDmytruk: This data is `int` in other places and I think it can be `static`.
```suggestion
static int txt_booting_apicid;
```

accek-itl: Fixed.

xen/arch/x86/boot/x86_64.S
accek-itl: Improved the comment, please have a look if this is good now.

SergiiDmytruk (approved):

PR #29, opened 2026-04-23
xen/arch/x86: add TPR (TXT Protected Range) DMA protection support

accek-itl:
Assisted-by: Claude:claude-opus-4-6

### DIFF
--- a/xen/arch/x86/boot/head.S
+++ b/xen/arch/x86/boot/head.S
@@ -136,12 +136,12 @@ SYM(mle_header, DATA, LOCAL, 16)
         .long   0xa2555c0f  /* UUID2 */
         .long   0x42b651cb  /* UUID3 */
         .long   (.Lmle_header_end - mle_header)  /* MLE header size */
-        .long   0x00020002  /* MLE version 2.2 */
+        .long   0x00020003  /* MLE version 2.3 */
         .long   (slaunch_stub_entry - start)  /* Linear entry point of MLE (SINIT virt. address) */
         .long   0x00000000  /* First valid page of MLE */
         .long   0x00000000  /* Offset within binary of first byte of MLE */
         .long   (_end - start)  /* Offset within binary of last byte + 1 of MLE */
-        .long   0x00000723  /* Bit vector of MLE-supported capabilities */
+        .long   0x00004723  /* Bit vector of MLE-supported capabilities */
         .long   0x00000000  /* Starting linear address of command line (unused) */
         .long   0x00000000  /* Ending linear address of command line (unused) */
 .Lmle_header_end:
--- a/xen/arch/x86/include/asm/intel-txt.h
+++ b/xen/arch/x86/include/asm/intel-txt.h
@@ -245,8 +245,8 @@ struct txt_ev_log_container_12 {
  */
 struct txt_ext_data_element {
     uint32_t type;   /* One of TXT_HEAP_EXTDATA_TYPE_*. */
-    uint32_t size;
-    uint8_t data[0]; /* size bytes. */
+    uint32_t size;   /* Size of the whole element (header + data), in bytes. */
+    uint8_t data[0];
 } __packed;
 
 /*
--- a/xen/arch/x86/include/asm/intel-txt.h
+++ b/xen/arch/x86/include/asm/intel-txt.h
@@ -65,6 +65,9 @@
 #define SLAUNCH_ERROR_NO_VENDOR_INFO    0xc0008009U
 #define SLAUNCH_ERROR_BAD_VENDOR_INFO   0xc000800AU
 #define SLAUNCH_ERROR_BAD_SLRT_ADDRESS  0xc000800BU
+#define SLAUNCH_ERROR_TPR_INVALID       0xc000800CU
+#define SLAUNCH_ERROR_TPR_UNSUPPORTED   0xc000800DU
+#define SLAUNCH_ERROR_TPR_NOT_FOUND     0xc000800EU
 
 #define TXT_AP_BOOT_CS                  0x0030
 #define TXT_AP_BOOT_DS                  0x0038
@@ -73,6 +76,9 @@
 #define GETSEC_CAPABILITIES             0
 /* Intel SDM: GETSEC Capability Result Encoding */
 #define GETSEC_CAP_TXT_CHIPSET          1
+
+/* SINIT/MLE capability bit for TPR (TXT Protected Range) DMA protection. */
+#define TXT_SINIT_MLE_CAP_TPR_SUPPORT   14
 
 #ifndef __ASSEMBLER__
 
@@ -260,6 +266,19 @@ struct heap_event_log_pointer_element2_1 {
 } __packed;
 
 /*
+ * Extended data describing TPR (TXT Protected Range) DMA protection ranges.
+ */
+struct txt_heap_tpr_range {
+    uint64_t base;
+    uint64_t size;
+} __packed;
+
+struct txt_heap_tpr_req_element {
+    uint32_t count;
+    struct txt_heap_tpr_range ranges[0];
+} __packed;
+
+/*
  * Functions to extract data from the Intel TXT Heap Memory.
  *
  * The layout of the heap is dictated by TXT. It's a set of variable-sized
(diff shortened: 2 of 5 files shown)


xen/arch/x86/include/asm/intel-txt.h
```diff
@@ -327,67 +346,153 @@ static inline void *txt_init(void)
     return txt_heap;
 }
 
-static inline bool is_in_pmr(const struct txt_os_sinit_data *os_sinit,
-                             uint64_t base, uint32_t size, bool check_high)
+/*
+ * Find the TPR request element in the TXT heap extended data.
+ */
+static inline const struct txt_heap_tpr_req_element *
+txt_find_tpr_req_element(const struct txt_os_sinit_data *os_sinit)
```
SergiiDmytruk: If you'll make this function take data element type as a parameter, it will be possible to reuse the function in `tpm.c` which is the only other place that searches for a data element.

P.S. Looks like `tpm.c` is actually doing it wrong: it interprets size relative to the `data` field which matches the comment on `txt_ext_data_element` in this file but contradicts the SDM.  So it shouldn't actually work there and likely doesn't.

xen/arch/x86/include/asm/intel-txt.h
```diff
@@ -327,67 +346,153 @@ static inline void *txt_init(void)
...
+
     /* Check for size overflow. */
     if ( base + size < base )
         txt_reset(SLAUNCH_ERROR_INTEGER_OVERFLOW);
 
+    if ( os_sinit->capabilities & (1u << TXT_SINIT_MLE_CAP_TPR_SUPPORT) )
+    {
+        const struct txt_heap_tpr_req_element *tpr_req;
+
+        tpr_req = txt_find_tpr_req_element(os_sinit);
```
SergiiDmytruk: Maybe add a comment that `txt_verify_dma_protection()` has already validated presence and contents of the element.

xen/arch/x86/include/asm/intel-txt.h
```diff
@@ -327,67 +346,153 @@ static inline void *txt_init(void)
...
+        lo_size = tpr_req->ranges[0].base + tpr_req->ranges[0].size;
+        if ( tpr_req->count >= 2 )
```
SergiiDmytruk:
```suggestion
        if ( tpr_req->count == 2 )
```
or
```suggestion
        if ( tpr_req->count > 1 )
```

xen/arch/x86/include/asm/intel-txt.h
```diff
@@ -327,67 +346,153 @@ static inline void *txt_init(void)
...
+        lo_size = os_sinit->vtd_pmr_lo_size;
+        hi_base = os_sinit->vtd_pmr_hi_base;
+        hi_size = os_sinit->vtd_pmr_hi_size;
+    }
+
     /*
-     * txt_verify_pmr_ranges() makes sure the low range always starts at 0, so
-     * its size is also end address.
+     * txt_verify_dma_protection() makes sure the low range always starts at
+     * 0, so its size is also end address.
```
SergiiDmytruk, line 411: This comment should be in the `else`-branch above and the variable name should be `lo_end` instead of `lo_size` to make sense for both branches.

xen/arch/x86/include/asm/intel-txt.h
```diff
@@ -327,67 +346,153 @@ static inline void *txt_init(void)
...
+            txt_reset(SLAUNCH_ERROR_TPR_INVALID);
+        if ( tpr_req->count > 2 )
+            txt_reset(SLAUNCH_ERROR_TPR_UNSUPPORTED);
+
+        /* Lo range must not exceed 4G. */
+        if ( tpr_req->ranges[0].base + tpr_req->ranges[0].size >
+             0x100000000ULL )
+            txt_reset(SLAUNCH_ERROR_TPR_INVALID);
+
+        if ( tpr_req->count >= 2 )
```
SergiiDmytruk:
```suggestion
        if ( tpr_req->count == 2 )
```
or
```suggestion
        if ( tpr_req->count > 1 )
```

xen/arch/x86/include/asm/intel-txt.h
```diff
@@ -327,67 +346,153 @@ static inline void *txt_init(void)
...
+    }
+
     /*
-     * txt_verify_pmr_ranges() makes sure the low range always starts at 0, so
-     * its size is also end address.
+     * txt_verify_dma_protection() makes sure the low range always starts at
+     * 0, so its size is also end address.
      */
-    if ( base + size <= os_sinit->vtd_pmr_lo_size )
+    if ( base + size <= lo_size )
```
SergiiDmytruk, line 413: This check makes sense only if `lo_base` is guaranteed to be zero.  But it can't be with TPR due to overlaps with MMIO or something else, right?  Then need to check lower bound too.

xen/arch/x86/boot/head.S
```diff
@@ -136,12 +136,12 @@ SYM(mle_header, DATA, LOCAL, 16)
         .long   0xa2555c0f  /* UUID2 */
         .long   0x42b651cb  /* UUID3 */
         .long   (.Lmle_header_end - mle_header)  /* MLE header size */
-        .long   0x00020002  /* MLE version 2.2 */
+        .long   0x00020003  /* MLE version 2.3 */
```
SergiiDmytruk, line 139: I hope this won't cause older SINITs to refuse processing this MLE.

xen/arch/x86/include/asm/intel-txt.h
```diff
@@ -327,67 +346,153 @@ static inline void *txt_init(void)
...
+                txt_reset(SLAUNCH_ERROR_TPR_INVALID);
+
+            /* Ranges must not overlap. */
+            if ( tpr_req->ranges[0].base + tpr_req->ranges[0].size >
+                 tpr_req->ranges[1].base )
+                txt_reset(SLAUNCH_ERROR_TPR_INVALID);
+
+            /* All regions accessed by 32b code must be below 4G. */
+            if ( tpr_req->ranges[1].base + tpr_req->ranges[1].size <=
+                 0x100000000ULL )
```
SergiiDmytruk, line 481: `/* Hi range must start at or above 4G. */` check above leaves no chance for this condition to be true unless an integer overflow occurs.

Speaking of integer overflows, PMR code checks for them.

xen/arch/x86/include/asm/intel-txt.h
```diff
@@ -327,67 +346,153 @@ static inline void *txt_init(void)
...
+
+        tpr_req = txt_find_tpr_req_element(os_sinit);
+        if ( tpr_req == NULL )
+            txt_reset(SLAUNCH_ERROR_TPR_NOT_FOUND);
+        if ( tpr_req->count < 1 )
+            txt_reset(SLAUNCH_ERROR_TPR_INVALID);
+        if ( tpr_req->count > 2 )
+            txt_reset(SLAUNCH_ERROR_TPR_UNSUPPORTED);
+
+        /* Lo range must not exceed 4G. */
```
SergiiDmytruk:
```suggestion
        /* Low range must not exceed 4G. */
```

xen/arch/x86/include/asm/intel-txt.h
accek-itl: Done.

xen/arch/x86/include/asm/intel-txt.h
accek-itl: Done.

xen/arch/x86/include/asm/intel-txt.h
accek-itl: Removed, mimicing PMR logic

xen/arch/x86/include/asm/intel-txt.h
accek-itl: Done.

xen/arch/x86/include/asm/intel-txt.h
accek-itl: I currently boot with lo TPR base hardcoded in grub to 0, just like it's for PMR path.
Tboot uses lowest region from e820, rounded down to 2M boundary, both for PMR and TPR, which in practice should be 0.
So I adjusted this code to assume and check that. Now both PMR and TPR branches have the same logic.
I think we can add non-zero base support later, if needed.
Do you know any reason/platform that needs non-zero lo_base?

xen/arch/x86/include/asm/intel-txt.h
accek-itl: Kept the assumption that lo_base is always 0, also in TPR flow, so keeping the comment here.
Adjusted the if-branch to just assign the size.

xen/arch/x86/include/asm/intel-txt.h
accek-itl: Nice catch; it works by accident, because the event log element is the first one. I split this commit into multiple ones to separately extract the common os_sinit element search and apply it in TPM case. Only then add TPR support.

xen/arch/x86/include/asm/intel-txt.h
accek-itl: Added.

xen/arch/x86/boot/head.S
accek-itl: Yeah, me too. There's only min_mle_hdr_ver specified by SDG, so technically anything newer should always be compatible.

xen/arch/x86/include/asm/intel-txt.h
SergiiDmytruk: > Do you know any reason/platform that needs non-zero lo_base?

No, some prohibition of an overlap is the only thing I could think of, but MMIO windows usually aren't this low in memory.

xen/arch/x86/include/asm/intel-txt.h
```diff
@@ -327,67 +346,174 @@ static inline void *txt_init(void)
...
+    uint64_t lo_size, hi_base, hi_size;
+
     /* Check for size overflow. */
     if ( base + size < base )
         txt_reset(SLAUNCH_ERROR_INTEGER_OVERFLOW);
 
+    if ( os_sinit->capabilities & (1u << TXT_SINIT_MLE_CAP_TPR_SUPPORT) )
+    {
+
+        /*
```
SergiiDmytruk: nit: an extra line and a trailing space
```suggestion
        /*
```

xen/arch/x86/tpm.c
```diff
@@ -702,23 +702,16 @@ find_evt_log_ext_data(struct tpm2_spec_id_event *evt_log)
...
-        if ( ext_data->type == TXT_HEAP_EXTDATA_TYPE_EVENT_LOG_POINTER2_1 )
-            break;
-        ext_data = (void *)&ext_data->data[ext_data->size];
-    }
-
-    if ( ext_data->type == TXT_HEAP_EXTDATA_TYPE_END )
-        return NULL;
-
-    return (struct heap_event_log_pointer_element2_1 *)ext_data->data;
+    return (struct heap_event_log_pointer_element2_1 *)(uintptr_t)ext_data->data;
```
SergiiDmytruk: I think I've never seen such use of `uintptr_t` before, `const` can be casted away without it.  However, upstream will not like this with or without `uintptr_t`.  `txt_find_ext_data_element()` will have to accept/return non-const data, propagating non-const `txt_os_sinit_data` up the call chain.  Don't think there is a nicer way around it.

xen/arch/x86/include/asm/intel-txt.h
accek-itl: Fixed

xen/arch/x86/tpm.c
accek-itl: Yeah, I initially tried with removing const, but I didn't like this spilled to callers in `slaunch-early.c` making os_sinit variables non-const. I agree that upstream would probably don't like that, so here's the non-const version.

accek-itl (commented):
Here's the updated change. Compile-tested.

SergiiDmytruk (approved):

